PR #2, opened 2017-01-10
js_special_data_types - Darin Estrin

darin-estrin:


js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>js_special_data_types</title>
+    <script>
+      var x = null;
+      var y;
+      var z = 'I\'m a string!';
```
collette-tamez: At this point in JS the type of quotes you use isn't really important but its good that you remembered to escape your character here. However using double quotes would eliminate the need to do so.

collette-tamez:
#LFZ2

PR #4, opened 2017-01-10
js_primitive_data_types - Charles Kim

ck1114:


js_primitive_data_types/index.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <script>
+        //Feature Set 1
+        var firstName = "Charles";
+        var lastName = "Kim";
+        var age = "32";
```

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,31 @@
...
+        //Feature Set 2
+        var output1 = firstName + " " + lastName + " " + age;
+        console.log(output1)
+
+        //Feature Set 3
+        var x = 11;
+        var y = 14;
+
+        var output2 = x + y;
+        var output3 = x + "" + y;
```

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,31 @@
...
+        var output4 = "The total of" + " " + x + " " + "and" + " " + y + " " + "is" + " " + (x + y);
```
collette-tamez: Remember that literal strings also include character spaces. Its can get a bit confusing when concatenating long text like this, but you don't need the extra space here. You can instead include them like this: 
` var output4 = "The total of " + x + " and " + y + " is " + (x+y);`

Notice the extra space around `and` and `is`.

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,31 @@
...
+
+        console.log(output2);
```

collette-tamez:
#LFZ2

PR #6, opened 2017-01-10
js special data types - Mark Crisostomo

mcrisostomo9:


js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
...
+        //fs2
+        var output1 = x > y;
+        var output2 = x == z;
+        var output3 = y !== z;
+        console.log("Output1: ", output1);
+        console.log("Output2: ", output2);
+        console.log("Output3: ", output3);
+
+        //fs3
+        var output4 = 100;
```
collette-tamez: While giving these variables new values kind of works, you already have three declared variables that you can use. We want to illustrate how those three datatypes differ. Remember that `number` is a primitive data type.

js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
+    <script>
+        //fs1
+        var x = null;
+        var y;
+        var z = "I am super hungry";
```
collette-tamez: #same

collette-tamez:
#LFZ1

PR #7, opened 2017-01-10
js special data types - Derrick Sebesta

dsebesta:


js_special_data_types/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <meta charset="UTF-8">
+    <title>Title</title>
+    <script>
+        var x = null;
+        var y;
+        var z = 'derrick';
+        var output1 = x == y;
+        var output2 = x == z;
+        var output3 = y == z;
+        console.log(`My output1 is ${output1}`);
```
collette-tamez: Impressive use of template literals, just be aware of compatibility issues when using the new features of Javascript. While the browsers all support template literals, android mobile does not. Different companies have different standards for what browsers/devices they support so you may have to resort to using something like Babel to handle incompatibilities.

http://kangax.github.io/compat-table/es6/#test-template_literals

collette-tamez:
#LFZ2

collette-tamez:
#LFZ2

js_special_data_types/index.html
dsebesta: Ok, thank you for the heads up!

PR #8, opened 2017-01-10
js special data types - Charles Kim

ck1114:


js_special_data_types/index.html
```diff
@@ -0,0 +1,33 @@
...
+       var x = null;
+       var y;
+       var z = "zed";
+
+       //Feature Set 2
+       var output1 = x == y;
+       var output2 = x == z;
+       var output3 = y == z;
+
+       console.log("My output1", output1);
```
iamleonhardt: Nice work here. Using the , instead of concatenating here is nice because it gives you the space between the string"My output2" and the result. This is also really helpful when using objects because with the , it will actually show you whats in the object as well, instead of simply printing the name of the object.

Another option I like for labeling console.logs is to use spaces and semicolons to differentiate more clearly.  
ie. -  console.log('output1 is : ', output1');  It helps on a quick glance, especially when there are a lot of things going on in the console.

js_special_data_types/index.html
```diff
@@ -0,0 +1,33 @@
...
+       console.log("My output2", output2);
+       console.log("My output3", output3);
+
+       //Feature Set 3
+       var output4 = "x is an" + " " + typeof x;
```
iamleonhardt: For this group of console logs you've concatenated and empty string in the middle.  A more simple way (and more common way) to achieve the same thing is to simply add the empty space at the end of your previous string.  Such as:  "x is an " + typeof x;

This is also often used at the beginning of strings as needed.  For example:
console.log("The " + animalName + " says " + animalSound + ".";

js_special_data_types/index.html
```diff
@@ -0,0 +1,33 @@
...
+       var output5 = "y is an" + " " + typeof y;
+       var output6 = "z is an" + " " + typeof z;
+
+       console.log(output4);
```
iamleonhardt: Try to get in the habit of always adding a label to your console.logs. It will really pay off in the end.
console.log('output5 is : ', output5);

PR #14, opened 2017-01-10
js special data types - Gwenever Pacifico

gpacifico:


js_special_data_types/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Special Data Types</title>
+        <script>
+            var x = null;
+            var y;
+            var z = "milk and cereal";
```
iamleonhardt: 👍 delicious.

js_special_data_types/index.html
```diff
@@ -0,0 +1,26 @@
...
+            var output1 = x == y;
+            var output2 = x == z;
+            var output3 = y == z;
+            console.log("Gwen output1", output1);
```
iamleonhardt: Nice job here.  The , is nice because you can see the actual result of the variable more than simply the printed out string of it.  This becomes even more useful when dealing with objects as you can see whats inside of them as well.  I find it useful to add a space and colon to the label simply to help differentiate the result from the label. For example, console.log('output1 is : ', output1);

I'm curious why your console.logs have your name in them :)

PR #16, opened 2017-01-10
js_specData_types - Michael Duong

tallergeese:


js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>JS Special Data Types</title>
+
+    <script>
+        var x = null;
+        var y;
+        var z = "any string";
```
joshuabyun: Simple, succinct. Another shortcut way to declare multiple variables is to use comma like below. This could be a preference issue. 

`var x = null,
       y,
       z = "any string";`

js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
...
+
+        var output1 = x == y;
+        var output2 = x == z;
+        var output3 = y == z;
+
+        console.log("Output 1: ", output1);
```
joshuabyun: Good use of comma when printing out instead of concatenating. This way, the output1 gets printed out fully if there are additional information such as prototype or attributes etc. The information provided can be very useful when you are debugging.

joshuabyun: By the way, what would you think if output1 was `x === y` ? would the answer have been the same or different? Something to think about.

js_special_data_types/index.html
```diff
@@ -0,0 +1,32 @@
...
+        console.log("Output 2: ", output2);
+        console.log("Output 3: ", output3);
+
+        var output4 = "x is a(n) " + typeof(x);
+        var output5 = "y is a(n) " + typeof(y);
+        var output6 = "z is a(n) " + typeof(z);
+
+        console.log("Output 4: ", output4);
```
joshuabyun: It is crucial to understand the difference between null and undefined, and how they act to different scenarios. They seem to be very similar and can get very tricky.

joshuabyun:
LF2

PR #18, opened 2017-01-10
js_primitive_data_types - Yaeri Lim

yaerilim:


js_primitive_data_types/index.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
+    <script>
+        var firstName = "Yaeri";
+        var lastName = "Lim";
+        var age = "28";
```

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,25 @@
...
+        var output1 = "First name: " + firstName + ", " + "Last Name: " + lastName +", " + "Age: " +age + "."
```
collette-tamez: Be very careful with your semi-colons. You want to make sure they're always there when they need to be and that things like braces and parenthesis are also closed properly. Syntax errors can be hard to spot and cause major headaches.

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,25 @@
...
+        console.log(output1)
+        var x = 12;
+        var y = 15;
+        var output2 = x + y;
+        var output3 = x + '' + y;
```

http://www.w3schools.com/jsref/jsref_tostring_number.asp

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,25 @@
...
+        var output4 = "The total of " + x + " and " + y + " is " + output2;
```
collette-tamez: Nice work using your `output2` here

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,25 @@
...
+        console.log(output2);
```

collette-tamez:
#LFZ2

PR #19, opened 2017-01-10
js special data types - Yaeri Lim

yaerilim:


joshuabyun:
Good job. Make sure to go over the difference "undefined" and "null". Null has type of object but undefined gives undefined type. Thus when you are comparing the 2 with '==', it will give true due to the value, but will give false when you use '===' because their types are different. 

We usually use null when we want to say "this variable has value ''nothing' ", very much like a place holder.

joshuabyun:
LF2

PR #21, opened 2017-01-10
js special data types - Andy Ong

Breathics:


js_special_data_types/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>JS Special Data Types</title>
+    <script>
+        var x = null;
+        var y;
+        var z = 'sandwich';
```
joshuabyun: Another shortcut way to declare multiple variables is to use comma like below. This could be a preference issue.

var x = null, y, z = "any string";

js_special_data_types/index.html
```diff
@@ -0,0 +1,28 @@
...
+
+        var output1 = x == y;
+        var output2 = x == z;
+        var output3 = y == z;
+        console.log("Output 1: " + output1);
```
joshuabyun: Did it come out what you expected? What would happen if you use '===' instead? Would the answer be the same?

js_special_data_types/index.html
```diff
@@ -0,0 +1,28 @@
...
+        console.log("Output 2: " + output2);
+        console.log("Output 3: " + output3);
+
+        var output4 = "x is an " + typeof x;
+        var output5 = "y is " + typeof y;
+        var output6 = "z is an " + typeof z;
+        console.log(output4);
```
joshuabyun: It is very important to understand the difference between null and undefined, and how they act to different scenarios. They seem to be very similar and can get very tricky.

joshuabyun:
LF2

PR #22, opened 2017-01-10
js_primitive_data_types - John Yosas

jayosas75:


js_primitive_data_types/index.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>JS Primitive Data Types</title>
+    <script>
+        // feature set 1
+        var firstName = 'John';
+        var lastName = 'Yosas'
+        var age = '27';
```

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,27 @@
...
+        var output1 = 'First Name: ' + firstName + ', ' + 'Last Name: ' + lastName + ', ' + 'Age: ' + age + '.';
+        console.log(output1);
+
+        //feature set 2
+        var x = 15;
+        var y = 21;
+        var output2 = x + y;
+        var output3 = x + '' + y;
```

http://www.w3schools.com/jsref/jsref_tostring_number.asp

js_primitive_data_types/index.html
```diff
@@ -0,0 +1,27 @@
...
+        var output4 = 'The total of ' + x + ' and ' + y + ' is ' + output2;
+
+        console.log(output2);
```

collette-tamez:
#LFZ2

PR #23, opened 2017-01-10
js spec data types - John Yosas

jayosas75:
Unsure why my Primitive data types index was including in this :O 

pls disregard that file. thanks

### DIFF
--- a/js_primitive_data_types/index.html
+++ b/js_primitive_data_types/index.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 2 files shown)


joshuabyun:
Good job. Make sure to go over the difference "undefined" and "null". Null has type object but undefined gives type undefined. When you are comparing the 2 with '==', it will give true due to the same value, but not when you use '===' because their types. Null is used when we want to say "this variable has value 'nothing' ", which can be used like sort of a placeholder. It will be very handy time to time. Keep up the good work!

joshuabyun:
Lf2

PR #27, opened 2017-01-10
Js comp cond - Yaeri Lim

yaerilim:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,62 @@
...
+        output2 = mySmallNumber === myString;
+        output3 = mySmallNumber < myBigNumber;
+        output4 = mySmallNumber > myBigNumber;
+        console.log("My output1 =", output1)
+        console.log("My output2 =", output2)
+        console.log("My output3 =", output3)
+        console.log("My output4 =", output4)
+        var gameNumber = 64;
+        var userInput1 = 12;
+        var userInput2 = gameNumber;
```
collette-tamez: Excellent work setting your variable here 👍

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,62 @@
...
+        if (userInput1 > gameNumber) {
+            console.log("number is too high");
+        }
+        if (userInput1 < gameNumber) {
+            console.log("number is too low");
+        }
```
collette-tamez: Good work using and `if` here instead of an `else if`

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,62 @@
...
+        if (1 > userInput1 > 100) {
```
collette-tamez: Be very careful here. While you're thinking in the right direction this conditional isn't doing what you think it is. What happens if you change your `userInput1 = 900`?  In this case you would be better off using `userInput1 < 1 || userInput1 > 100`

collette-tamez:
#LFZ2

PR #29, opened 2017-01-10
js comparisons and conditionals - Mark Crisostomo

mcrisostomo9:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+
+        //fs3
+        var gameNumber = 64;
+        var userInput1 = 99;
+        var userInput2 = gameNumber;
+
+        //fs4
+        if(userInput1 > gameNumber){
+            console.log("This number is too high!!!");
+        }
```
mcrisostomo9: There is no need to end this if statement with a semicolon right? I read that I shouldn't put a semicolon after curly brackets.

collette-tamez: @mcrisostomo9 That's correct, bracketed statements like this don't need a semicolon.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+        var myString = "5";
+        var myBigNumber = 255;
+
+        //fs2
+        var output1 = mySmallNumber == myString;
+        var output2 = mySmallNumber === myString;
+        var output3 = mySmallNumber < myBigNumber;
+        var output4 = mySmallNumber > myBigNumber;
+
+        console.log("Output 1: 5 is equal to \"5\" in value =", output1);
```
collette-tamez: Nice work escaping your quotes here. Remember that nested quotes are your friend:
`"Output 1: 5 is equal to '5' in value ="`
`'Output 1: 5 is equal to "5" in value ='`

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+        console.log("Output 2: 5 is equal to \"5\" in value and type =", output2);
+        console.log("Output 3: 5 is less than 255 =", output3);
+        console.log("Output 4: 5 is greater than 255 =", output4);
+
+        //fs3
+        var gameNumber = 64;
+        var userInput1 = 99;
+        var userInput2 = gameNumber;
```
collette-tamez: Nice work setting your variable here

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+
+        //fs4
+        if(userInput1 > gameNumber){
+            console.log("This number is too high!!!");
+        }
+        if(userInput1 < gameNumber) {
```

collette-tamez:
#LFZ2

PR #30, opened 2017-01-10
js comparisons and conditionals - Andy Ong

Breathics:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,70 @@
...
+        var output3 = mySmallNumber < myBigNumber;
+        var output4 = mySmallNumber > myBigNumber;
+        console.log("Output 1: "  + output1);
+        console.log("Output 2: "  + output2);
+        console.log("Output 4: "  + output4);
+
+        // Feature Set 3 - Declare More Variables
+        var gameNumber = 64;
+        var userInput1 = 35;
+        var userInput2 = gameNumber;
```
collette-tamez: Excellent work here setting your variable.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,70 @@
...
+
+        // Feature Set 4 - Conditionals
+        if (userInput1 > gameNumber){ //Condition 1
+            console.log("The inputted number is too high"); // true
+        }
+        else if (userInput1 < gameNumber){ //Condition 2
```
collette-tamez: If your `userInput1` is too high will your `else if` run? For the purposes of this assignment we want to check for for both cases of `userInput1` being too low and too high. The way your statement is set up now its checking for one or the other so you would be better of using two if statements in this case.

collette-tamez:
#LFZ2

PR #34, opened 2017-01-10
Js comp cond Jorge Cruz

RobzDoom:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,67 @@
...
+        console.log("My output =", output4);
+
+        var gameNumber = 64;
+        var userInput1 = 42;
+        var userInput2 = gameNumber;
+
+        if(userInput1 > gameNumber){
+            console.log('Number is too high')
+        }
+        else if(userInput1 < gameNumber){
```
collette-tamez: If your `userInput1` is too high will your `else if` run? For the purposes of this assignment we want to check for for both cases of `userInput1` being too low and too high. The way your statement is set up now its checking for one or the other so you would be better of using two if statements in this case.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,67 @@
...
+        var output4 = mySmallNumber > myBigNumber;
+
+        console.log("My output =", output1);
+        console.log("My output =", output2);
+        console.log("My output =", output3);
+        console.log("My output =", output4);
+
+        var gameNumber = 64;
+        var userInput1 = 42;
+        var userInput2 = gameNumber;
```
collette-tamez: Excellent work setting your variable here

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if(userInput2 == gameNumber){
+            console.log('The Correct Number has been guessed!')
+        }else{
+            console.log('Try again')
+        }
+
+        switch (userInput1){
+            case (gameNumber):
```
collette-tamez: I'm sure it makes it easier to read but you can drop the parentheses here as they're not really part of the syntax for switch statements.

collette-tamez:
#LFZ2

PR #36, opened 2017-01-10
js comparisons and conditionals - John Yosas

jayosas75:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+        var output4 = mySmallNumber > myBigNumber;
+        console.log('My output1 = ' + output1);
+        console.log('My output2 = ' + output2);
+        console.log('My output3 = ' + output3);
+        console.log('My output4 = ' + output4);
+
+        //feature set 3 and 4
+        var gameNumber = 64;
+        var userInput1 = 88;
+        var userInput2 = 64;
```

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+        if (userInput1 > gameNumber) {
+            console.log('the number is too high');
+        }
+        if (userInput1 < gameNumber) {
```

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+            console.log('the number is in range');
+        }
+        if (userInput2 === gameNumber) {
+            console.log('the correct number has been guessed');
+        } else {
+            console.log('please try again')
+        }
+
+        //feature set 5
+        switch(true) {
```
collette-tamez: So lets review switch statements 
http://www.w3schools.com/jsref/jsref_switch.asp
The parameter you pass into your switch is what each `case` compares against. So although it looks like this worked, we instead we only want to pass in our `userInput1` since thats what we to compare against each case. Which means we want to use `switch(userInput1)`

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+            case userInput1 === gameNumber:
+                console.log('You found the correct number!');
+                break;
+            case userInput1 < gameNumber && gameNumber - 10 <= userInput1:
+                console.log('Too low but warm.');
+                break;
+            case userInput1 > gameNumber && gameNumber + 10 >= userInput1:
```
collette-tamez: With switch statements we really don't need to write full conditionals like this. The switch statement should already compare `userInput1` to whatever our case statement is. So rather than write this long conditional we can instead use 
```Javascript
case gameNumber + 10:
     \\\code to execute
     break;
```

collette-tamez:
#LFZ1

PR #38, opened 2017-01-10
js functions - Andy Ong

Breathics:


js_functions/main.js
```diff
@@ -0,0 +1,21 @@
...
+    console.log("It is raining sideways!");
+}
+
+function add(x, y){
+    var total = x + y;
+    console.log("Your total will be ... " + total);
+}
+
+function add2(x, y){
+    var total = x + y;
```
collette-tamez: Excellent work on assigning your total to a variable in both your functions here.

js_functions/main.js
```diff
@@ -0,0 +1,21 @@
...
+    return total;
+}
+var add2result = add2(6, 12);
+
+function cardFlip(hideClicked){
+    $(hideClicked).hide();
```
collette-tamez: Nice work using `.hide()`, theres usually more than one way to approach a problem in jQuery so be sure to experiment the different methods. I find that `.toggle()` is prefect for achieving a card flip.

PR #39, opened 2017-01-10
js functions - John Yosas

jayosas75:


js_functions/main.js
```diff
@@ -0,0 +1,21 @@
...
+function myMessage(){
+    console.log('Can\'t spell function without fun.');
+}
+function add(x,y) {
+    console.log(x + y);
+}
+
+//feature set 2
+function add2(x,y) {
+    var total = x + y;
```
collette-tamez: Nice work setting up your variable here. I would definitely suggest implementing it in your `add`function as its a good habit to get into.

PR #40, opened 2017-01-10
js comparisons and conditionals - Derrick Sebesta

dsebesta:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+        var output3 = mySmallNumber < myBigNumber;
+        var output4 = mySmallNumber > myBigNumber;
+        console.log('***** Feature Set 2 *****');
+        console.log(`Is ${mySmallNumber} equal to ${myString}: ${output1}`);
+        console.log(`Is the value and datatype of ${mySmallNumber} equal to ${myString}: ${output2}`);
+        console.log(`Is ${mySmallNumber} less than ${myBigNumber}: ${output3}`);
+        console.log(`Is ${mySmallNumber} greater than ${myBigNumber}: ${output4}`);
+        var gameNumber = 31;
+        var userInput1 = 56;
+        var userInput2 = 54;
```
collette-tamez: You want to set `userInput2` to your game number. The easiest way to approach that would be `userInput2 = gameNumber`. Anytime you wanted to change the value of `gameNumber` you would only have to change the original value on line 29. That way you don't have to spend your time searching through code for places where the values are the same.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,65 @@
...
+        console.log('***** Feature Set 4 *****');
+        if (userInput1 > gameNumber) {
+            console.log(`Your number (${userInput1}) is too damn high!`)
+        }
+        if (userInput1 < gameNumber) {
```

collette-tamez:
#LFZ2

PR #44, opened 2017-01-10
js comparisons and conditionals - Charles Kim

ck1114:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+        console.log("My output =", output1);
+        console.log("My output =", output2);
+        console.log("My output =", output3);
+        console.log("My output =", output4);
+
+        //Feature Set 3
+        var gameNumber = 64;
+        var userInput1 = 14;
+        var userInput2 = 64;
```
collette-tamez: Here it would be more efficient if you set `var userInput2 = gameNumber;`. Anytime you wanted to change the value of `gameNumber` you would only have to change the original value on line 29. That way you don't have to spend your time searching through code for places where the values are the same.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+        //Feature Set 4
+        if (userInput1 > gameNumber){
+            console.log("the number is too high");
+        }
+
+        if (userInput1 < gameNumber){
```
collette-tamez: Nice work using two `if` statements instead of an `else` here since we wanted you to make both comparisons.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,77 @@
...
+                break;
+
+            case gameNumber - 25:
+                console.log("Too low and cold");
+                break;
+
+            case gameNumber + 25:
+                console.log("Too high and cold");
+                break;
+
```
collette-tamez: I would recommend eliminating your empty lines and using comments to keep yourself organized (though you will want to remove them later). As your javascript files get longer with the more code you write, you want to able to make them as easy to read as possible.

collette-tamez:
#LFZ2

PR #45, opened 2017-01-10
js_comparison_and_conditionals - Darin Estrin

darin-estrin:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,75 @@
...
+
+        console.log("My outPut1 = " + outPut1);
+        console.log("My outPut2 = " + outPut2);
+        console.log("My outPut3 = " + outPut3);
+        console.log("My outPut4 = " + outPut4);
+
+        // Feature Set 3
+        gameNumber = 64;
+        userInput1 = 20;
+        userInput2 = 64;
```
collette-tamez: Here it would be more efficient if you set `var userInput2 = gameNumber;`. Anytime you wanted to change the value of `gameNumber` you would only have to change the original value on line 29. That way you don't have to spend your time searching through code for places where the values are the same.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,75 @@
...
+
+        // Feature Set 4
+        if (userInput1 > gameNumber){
+            console.log('Your number is too high.');
+        }
+
+        if (userInput1 < gameNumber){
```
collette-tamez: Nice work using two `if` statements instead of an `else` here since we wanted you to make both comparisons.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        if(userInput2 === gameNumber){
+            console.log('You guessed the correct number');
+        } else {
+            console.log('Try again.')
+        }
+
+        // Feature Set 5
+        switch(true){
```
collette-tamez: So lets review switch statements 
http://www.w3schools.com/jsref/jsref_switch.asp
The parameter you pass into your switch is what each `case` compares against. So although it looks like this worked, we instead we only want to pass in our `userInput1` since thats what we to compare against each case. Which means we want to use `switch(userInput1)`

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,75 @@
...
+            case userInput1 === gameNumber:
+                console.log('You found the correct number!');
+                break;
+            case userInput1 < gameNumber && userInput1 >= gameNumber - 10:
```
collette-tamez: With switch statements we really don't need to write full conditionals like this. The switch statement should already compare `userInput1` to whatever our case statement is. So rather than write this long conditional we can instead use 
```Javascript
case gameNumber + 10:
     \\\code to execute
     break;
```

collette-tamez:
#LFZ1

PR #46, opened 2017-01-10
js_dom_events - Darin Estrin

darin-estrin:


darin-estrin:
So I don't know why is says there are two participants, but bonehenge was my first github account. I made a new one since I thought using the other one wouldn't appear as professional.

collette-tamez:
@darin-estrin did you log into that github account when setting up phpstorm? I had to hazard a guess theres probably a flag set somewhere on your system with that login info.

collette-tamez:
#LFZ2

PR #49, opened 2017-01-10
js comparisons and conditionals - Michael Duong

tallergeese:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+        console.log("Output 1: ", output1);
+        console.log("Output 2: ", output2);
+        console.log("Output 3: ", output3);
+        console.log("Output 4: ", output4);
+
+        //Feature Set 3
+        var gameNumber = 64;
+        var userInput1 = 1;
+        var userInput2 = 64;
```

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+        //Feature Set 4;
+
+        if (userInput1 > gameNumber) {
+            console.log("Input 1 is too high.")
+        }
+
+        if (userInput1 < gameNumber) {
```

collette-tamez:
#LFZ2

PR #55, opened 2017-01-11
Js functions - Mark Crisostomo

mcrisostomo9:


js_functions/main.js
```diff
@@ -0,0 +1,22 @@
...
+//fs1.3
+function add(x, y){
+    var total = x + y;
+    console.log(total);
+}
+
+//fs2.1
+function add2(x, y){
+    var total = x + y;
+    return total;
```
collette-tamez: Excellent work setting up you variables on your two functions. Its habit you definitely want to get into and utilize whenever possible.

PR #58, opened 2017-01-11
js comparisons and conditionals - brock patterson

brockcp:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,115 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <script>
+        //feature set 1
+        var mySmallNumber = 5;
+        var myString = "5";
+        var myBigNumber = 255;
+
```
collette-tamez: I would recommend eliminating your empty lines. As your javascript files get longer with the more code you write, you want to able to make them as easy to read as possible.

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,115 @@
...
+        var output1 = mySmallNumber == myString;
+        console.log("output1 is " + output1);
+
+        //output2
+        var output2 = mySmallNumber === myString;
+        console.log("output2 is " + output2);
+
+        //output3
+        var output3 = mySmallNumber < myBigNumber;
+        console.log("output3 is " + output3);
```
collette-tamez: Nice work grouping these together 👍

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,115 @@
...
+
+        //or..?
+            var output4b = (mySmallNumber > myBigNumber) ? "true" : "false";
+            console.log("output4b is " + output4b);
+
+
+        //feature set 3
+        var gameNumber = 64;
+        var userInput1 = 74;
+        var userInput2 = gameNumber;
```
collette-tamez: Nice work setting your variable here 👍

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,115 @@
...
+        //if#2
+        if(userInput1 < gameNumber){
+            console.log("the number is too low");
+        }
+        else{
+            console.log("the number is not too low")
+        }
+
+        //if#3
+        if(userInput1 < 1 && userInput1 > 100){
```
collette-tamez: Be careful with your logical operators.  `&&` is not only checking that both conidtionals are true but also checks that _both conditions are true at the same time_. Your if statement here will always log your else statement since `userInput1` can't be both less and than zero and greater than 100 at the same time. Instead you'll want to you use the "or" operator: `||` which is only needs one of the conditionals to be true.

`(userInput1 < 1 || userInput1 > 100)`

collette-tamez:
#LFZ2

PR #60, opened 2017-01-11
js comparisons and conditionals - Gwenever Pacifico

gpacifico:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,64 @@
...
+            var output3 = mySmallNumber < myBigNumber;
+            var output4 = mySmallNumber > myBigNumber;
+            console.log("My output =", output1);
+            console.log("My output =", output2);
+            console.log("My output =", output3);
+            console.log("My output =", output4);
+
+            var gameNumber = 64;
+            var userInput1 = 42;
+            var userInput2 = 64;
```

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,64 @@
...
+            if (userInput1 > gameNumber) {
+                console.log("Sorry Goldilocks, that number is too high.");
+            }
+            if (userInput1 < gameNumber) {
```

collette-tamez:
#LFZ2

PR #61, opened 2017-01-11
js functions - Tim Feng

spikers:


js_functions/main.js
```diff
@@ -0,0 +1,17 @@
+function myMessage() {
+  console.log("This is Tim's custom message!")
+}
+
+function add(firstNumber, secondNumber) {
+  console.log(firstNumber + secondNumber);
```
collette-tamez: In this function and your `add2` it would be more efficient to save your result to a variable and then console log that value. It would make labeling your console log a little easier as well.
```Javascript
function add(num1, num2){
     var result = num1 + num2;
     console.log("My function result is: " + result);
}
////OR
function add2(num1, num2){
     var secondResult = num1 + num2;
     return secondResult;
}
```

js_functions/main.js
```diff
@@ -0,0 +1,17 @@
+function myMessage() {
+  console.log("This is Tim's custom message!")
```
collette-tamez: Be careful with your semi-colons and syntax in general. At best nothing will change and at worst something will break and they can be tricky errors to spot.

js_functions/main.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+var add2Result = add2(10, 55);
+
+function add2(firstNumber, secondNumber) {
+  return firstNumber + secondNumber;
+}
+
+function cardFlip(element) {
+  element.style.display = "none";
```
collette-tamez: Excellent work using the vanilla JavaScript method.  Do take some time to play around with jQuery as it pretty versatile. For instance there is a `toggle()` method in jQuery that is super useful in achieving a "card flip".

collette-tamez:
#LFZ2

PR #62, opened 2017-01-11
js jquery - Tim Feng

spikers:


js_jquery/index.html
```diff
@@ -31,13 +32,35 @@
...
                     console.log("I am White!");
+                    
+                    //Doing this in jQuery//
+                    $("#div2").text("I am White!");
                 }
             });
             //----Write your code bellow----//
-
+            //Gonna use this for vanilla javascript only//
+            var div3Selector = document.querySelector("#div3");
```
collette-tamez: Assigning a selector to a variable is a very wise choice. Its definitely a habit you want to carry into jQuery. Another best practice to include with that is starting  selectors assigned to variables like this with a `$`
`var $div3Selector = document.querySelector("#div3");`  

The dollar sign itself isn't really a special character so its valid for variable names.

js_jquery/index.html
```diff
@@ -31,13 +32,35 @@
...
+            
+            function classToggler(selectedElement) {
+              selectedElement.className == "card-back" ? selectedElement.className += " flag" : selectedElement.className = "card-back";
+            }
+            
+            //Gonna use this for jQuery only//
+            $("#div4").addClass("card-back");
+            $("#div4").click(function() {
+              $(this).toggleClass("path");
+            });
```
collette-tamez: Nice work setting up these ones.

If you haven't already be sure to try chaining your methods together. Its something you'll definitely want to have a good grasp on in the future.  

You can also look at jQuery selectors for a different approach with these two lines.
http://www.w3schools.com/jquery/jquery_ref_selectors.asp

js_jquery/index.html
spikers: Good tip! I'll start doing this from week 2 onwards.

PR #64, opened 2017-01-11
Js jquery Jorge Cruz

RobzDoom:


js_jquery/index.html
```diff
@@ -14,13 +14,13 @@
...
-            $("#div1").addClass("red"); //change the word red to pink *note these classes have already been created in the CSS file
+            $("#div1").addClass("pink"); //change the word red to pink *note these classes have already been created in the CSS file
 
             //.click is like using the onclick attribute we used before
             //It creates an event handler for when div1 is clicked
             $("#div1").click(function(){
-                console.log("You clicked div 1, It is red"); //Fix the output to reflect div1's new color
-
+                console.log("You clicked div 1, It is pink"); //Fix the output to reflect div1's new color
+                $('#div1').html('Click Counts ' + count++)
```
collette-tamez: Pay close attention to what value the click counter starts at. Should it start at zero even though you've already clicked it once? 

Where you add the `++` actually changes when your `count` variable increments. Try placing the `++` before your count variable and notice how things change.

You can also take an alternate approach and increment outside of your console log.
```javascript
$("#div1").click(function(){
     console.log("You clicked div 1, It is pink");
     count++;
     $(this).html('Click Counts ' + count);
}
```
You already have a variable set to store the number of clicks, so like your functions prototype you can just log `count`.

js_jquery/index.html
```diff
@@ -31,15 +31,23 @@
                 if($("#div2").hasClass("green")){
                     //If it has the class "green" then it will output "I am Green!"
                     console.log("I am Green!");
+                    $('#div2').html('I am color green')
```
collette-tamez: Make sure you understand the differences between `.html()` and `.text()`. They differ in their functionality and one leaves you open to certain vulnerabilities.

js_jquery/index.html
```diff
@@ -31,15 +31,23 @@
...
                 }else{
                     //else it does not have the class "green" then it will output "I am White!"
                     console.log("I am White!");
+                    $('#div2').html('I am color white')
                 }
             });
             //----Write your code bellow----//
+            $('#div2').addClass('card-back');
```
collette-tamez: You're applying the class the wrong div here. \
screen shot 2017-01-13 at 11 55 20 am
Always check your work as you're working on it and then always double check the prototype instructions.

PR #65, opened 2017-01-11
js_jquery - Darin Estrin

darin-estrin:


js_jquery/index.html
```diff
@@ -8,18 +8,19 @@
     <!-- This should always be before your script -->
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js"></script>
     <script>
-        var count = 0; // A counter variable to be used in Feature Set 3
+        var count = 1; // A counter variable to be used in Feature Set 3
```
collette-tamez: I see your thinking here, but really we don't need to change the counter here. Think about the what the counter represents. The `count` variable here is meant to keep track of the number of times the user clicks on `div1`. Once the page finishes loading, has the user clicked on anything yet?

js_jquery/index.html
```diff
@@ -8,18 +8,19 @@
...
             //here we are targeting the div1 element
-            $("#div1").addClass("red"); //change the word red to pink *note these classes have already been created in the CSS file
+            $("#div1").addClass("pink"); //change the word red to pink *note these classes have already been created in the CSS file
 
             //.click is like using the onclick attribute we used before
             //It creates an event handler for when div1 is clicked
             $("#div1").click(function(){
-                console.log("You clicked div 1, It is red"); //Fix the output to reflect div1's new color
+                console.log("You clicked div 1, It is now pink"); //Fix the output to reflect div1's new color
+                $(this).html('Click Count: ' + count++);
```
collette-tamez: Where you add the `++` actually changes when your `count` variable increments. Try placing the `++` before your count variable and notice how things change.

You can also take an alternate approach and increment outside of your console log.
```javascript
$("#div1").click(function(){
     console.log("You clicked div 1, It is pink");
     count++;
     $(this).html('Click Counts ' + count);
}
```
So like your functions prototype you can just log `count` and now theres no need to start `count` at 1

js_jquery/index.html
```diff
@@ -30,15 +31,21 @@
                 //Here is an if statement that outputs the color of div2 to the console when it is clicked
                 if($("#div2").hasClass("green")){
                     //If it has the class "green" then it will output "I am Green!"
-                    console.log("I am Green!");
+                    $(this).html("I am Green!");
```
collette-tamez: Make sure you understand the differences between `.html()` and `.text()`. They differ in their functionality and one leaves you open to certain vulnerabilities.

js_jquery/index.html
```diff
@@ -30,15 +31,21 @@
...
                 }else{
                     //else it does not have the class "green" then it will output "I am White!"
-                    console.log("I am White!");
+                    $(this).html("I am White!");
                 }
             });
             //----Write your code bellow----//
+            $('#div3, #div4').addClass('card-back');
```
collette-tamez: Excellent work making use of jQuery selectors here 😎  Much like the CSS selectors, theres quite a few of them and can be super versatile.

PR #66, opened 2017-01-11
js_functions - Darin Estrin

darin-estrin:


js_functions/main.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+function add(a, b) {
+    console.log(a + b);
+}
+
+
+// Feature Set 2
+function add2(x, y) {
+    var total = x + y;
```
collette-tamez: Excellent work here using a separate variable 🥇  Its definitely a more efficient approach.

js_functions/main.js
```diff
@@ -0,0 +1,22 @@
...
+    return total;
+}
+
+var add2result = add2(10, 36);
+
+// Feature Set 3
+function cardFlip(element) {
+    $(element).hide();
```

collette-tamez:
#LFZ2

PR #69, opened 2017-01-11
js functions - Gwenever Pacifico

gpacifico:


js_functions/main.js
```diff
@@ -0,0 +1,18 @@
...
+
+function add2(x, y) {
+    var total = x + y;
+    return total;
+}
+
+var add2result = add2(10, 36);
+
+function cardFlip(element) {
+    $(element).hide();
```

PR #71, opened 2017-01-11
js functions - Charles Kim

ck1114:


ck1114:
Sorry. It also pulled my js_query prototype. I am a little unclear how I commit not every change I made, but only certain files.

js_functions/main.js
```diff
@@ -0,0 +1,18 @@
+function myMessage(){
+    console.log("This is my first fuction!");
+}
+
+function add(x,y){
+    console.log(x+y);
+}
+
+function add2(x,y){
+    var total = x + y;
```
collette-tamez: Excellent work here using a separate variable. Its a more efficient approach, and I would suggest implementing it in your `add` function.

js_functions/main.js
```diff
@@ -0,0 +1,18 @@
...
+    return total;
+}
+
+var add2result = add2(10,40);
+
+function cardFlip(element){
+    $(element).hide();
```

collette-tamez:
#LFZ2

PR #72, opened 2017-01-12
js dom creation - Andy Ong

Breathics:


js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
     <script>
         $(document).ready(function () {
             //playground 1 code
+            var pGround1Heading = $("<h1>I am a heading</h1>");
+            var pGround1Paragraph = $("<p>").text('This is an awesome paragraph');
+            var redStyling = {
+                text: 'This has red text',
+                style: 'color:red'
+            };
+            var pGround1Red = $("<p>", redStyling);
```
collette-tamez: Excellent work using the different approaches for creating an element in jQuery.

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
...
+
+            $('#playground1.playground_destination').append(pGround1Heading, pGround1Paragraph, pGround1Red)
```
collette-tamez: Your div here already has an id, theres not really a need to include the class in your selector. The `.removeClass()` method will still work without it. 

Great work using `.append()` to apply all your child elements at once.

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
...
+            var pGroundDiv3 = $("<div class='awesome_class'>Here is <strong>EVEN MORE</strong> awesome div text</div>");
+            $('#playground2.playground_destination').append(pGroundDiv1, pGroundDiv2, pGroundDiv3)
+                .removeClass('playground_destination').addClass('playground_source');
 
             //playground 3 code
+            var linksList = $('<ul>').addClass('neato');
+            for (var i=1; i<4; i++) {
+                var pageLink = $('<li>').text('Link ' + i);
+                linksList.append(pageLink);
+            }
```
collette-tamez: Excellent work using a for loop here 👏  👏

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
...
+
+            var pGroundFigure = $('<figure>').addClass('awesome_picture')
+            var awesome_doge = $('<img>').attr('src', 'http://comerecommended.com/wp-content/uploads/201' +
+                '3/01/how-to-get-awesome-back.jpg');
```
collette-tamez: Nice work using `.attr()`, this would be one way to approach adding your `data-coolness='high'` attribute to your `pGroundDiv3`.

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
...
+            var dog_caption = $("<figcaption>", {
+                text: "That's right, ",
+                append: $("<span>", {
+                    text: "you are",
+                }),
```
collette-tamez: Very nice work using an object here. Personally I prefer this approach when applying attributes, especially to image tags.

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
...
+            });
+            pGroundFigure.append(awesome_doge, dog_caption);
+            $('#playground3.playground_destination').append(linksList, pGroundFigure)
+                .removeClass('playground_destination').addClass('playground_source');
```
collette-tamez: Bonus points

js_dom_creation/index.html
```diff
@@ -41,10 +41,43 @@
     <script>
         $(document).ready(function () {
             //playground 1 code
+            var pGround1Heading = $("<h1>I am a heading</h1>");
```
Breathics: I'm assuming this is for increased readability/context for both me and other coders? I'm trying to make the most sense of this haha

collette-tamez: Yeah its more a readability thing and by no means a requirement. Like think about how the jQuery selectors start with a dollar sign. If you start your own variable with a dollar sign its a quick way to indicate that its value is a dom element.
```Javascript
var $element = $("element");
///so for example
var $responseParagraph = $("<p>");
$responseParagraph.text("You guessed it!");
$("#response_div").append($responseParagraph);
```

PR #76, opened 2017-01-12
js functions - Daniel Fogerty

DanFog:


js_functions/main.js
```diff
@@ -0,0 +1,18 @@
+function myMessage () {
+    console.log("This is my first function! Woo!");
+}
+
+function add(x, y) {
+    console.log(x + y);
+}
+
+function add2(x, y) {
+    var total = x + y;
```
collette-tamez: Excellent work here using a separate variable. Its a more efficient approach, and I would suggest implementing it in your `add` function.

js_functions/main.js
```diff
@@ -0,0 +1,18 @@
...
+    return total;
+}
+
+var add2result =add2(5, 12);
+
+function cardFlip(x) {
+    $(x).hide();
```

Its also important have clear and concise naming conventions. While the use of `x` and `y` can make sense from a math standpoint, its a little more vague the way you use it here. I would use `element` instead. 
```Javascript
function cardFlip(element){
     $(element).hide();
}
```
Using `element` tells you exactly whats happening in a simple way. We know that an element in our document is going to passed into the funtion, and we know that we're taking that dom element and wrapping it in a jQuery selector.

collette-tamez:
#LFZ2

PR #77, opened 2017-01-12
Js dom creation

spikers:


js_dom_creation/index.html
```diff
@@ -41,10 +41,49 @@
     <script>
         $(document).ready(function () {
             //playground 1 code
+            $("#playground1").html("<h1>I am a heading</h1><p>This is an awesome paragraph</p><p style='color:red'>This has red text</p>");
 
             //playground 2 code
+            $("#playground2").html("<div class='awesome_class'>Here is some awesome div text</div><div class='awesome_class' id='my_awesome_div2'>Here is MORE awesome div text</div><div class='awesome_class' id='my_awesome_div3' data-coolness='high'>Here is <strong>EVEN MORE</strong> awesome div text</div>");
```
collette-tamez: Its good that you're familiar with the vanilla Javascript methods. It certainly illustrates the differences between something quick like this vs. a piece by piece method. That being said, you definitely want to try the different methods of dynamically creating elements in jQuery. I'm quite partial to the object method.

PR #79, opened 2017-01-12
js functions - brock patterson

brockcp:


js_functions/main.js
```diff
@@ -0,0 +1,18 @@
+function myMessage (){
+    console.log("this is my first function")
+}
+
+function add (a,b){
+    console.log(a + b);
+}
+
+function add2 (a,b){
+    var add2_total = a + b;
```
collette-tamez: Excellent work here using a separate variable. Its a more efficient approach, and I would suggest implementing this approach in your `add` function.

js_functions/main.js
```diff
@@ -0,0 +1,18 @@
...
+    return add2_total;
+}
+var add2result = add2(10,5);
+
+function cardFlip(dontknowthis){
```
collette-tamez: Its important to have clear and concise naming conventions especially when sharing code. I would use `element` instead. 
```Javascript
function cardFlip(element){
     $(element).hide();
}
```
Using `element` tells you exactly whats happening in a simple way. We know that an element in our document is going to passed into the function, and we know that we're taking that dom element and wrapping it in a jQuery selector.

js_functions/main.js
```diff
@@ -0,0 +1,18 @@
...
+    $(dontknowthis).hide();
```
collette-tamez: There's certainly more than one way to approach things in jQuery so definitely take some time to play around with the different methods you find. For instance jQuery has `.toggle()` method that would be a great way for achieving a "card flip".

collette-tamez:
#LFZ2

PR #82, opened 2017-01-12
js repeat blocks - Tim Feng

spikers:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <script>
+    //Feature Set 1
+    var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    
+    logArray(numbers);
```
joshuabyun: It's good to see you passing in function as a some what callback, but try to write your code simple when you need to. It seems like your code is more complicated than what it should be. something like `numbers.forEach(function(value){console.log(value)});` would do the same. I think good code is when you write something that could be easily be read and understood by others.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+      console.log("");
+    }
+    
+    function logItem(item) {
+      console.log(item);
+    }
+
+    logArray(arrayReverser(numbers));
+    
+    //I did it this way because I realized array.reverse was destructive.
```
joshuabyun: Agree, array.reverse would change the original array as well. For loop works well in this case! One suggestion you can try is using `var newArray = numbers.slice()`, .slice() being array method. This would return a new set of array separate from your `numbers` array.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+      return tempArray;
+    }
+
+    logArray(extractOddNumbers(numbers));
+    
+    function extractOddNumbers(arr) {
+      var tempArray = [],
+          counter = 0;
+      for (var i = 0; i < arr.length; i++) {
+        if (arr[i] % 2 !== 0) {
```
joshuabyun: Good use of modulo!

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+      }
+      return tempArray;
+    }
+    
+    var mixedArray = [4, 56, 11, 13, 97, 22, 7, 36, 19, 5, 8, 10];
+    logArray(extractOddNumbers(mixedArray));
+    
+    //Feature Set 2: PLEASE GO TO LINE 98 FOR FOR-IN SOLUTION
+    var classRoom = {desk1: 'Stu', desk2: 'Peggy', desk3: 'LoneStar', desk4: 'George', desk5: 'Sheldon'};
+    var actionString = " is at ";
```
joshuabyun: I like how you set up a new variable for anything that could be used multiple times. Good practice!

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+    
+    loopItAllTogether(classRoom);
+    
+    function loopItAllTogether(obj) {
```
joshuabyun: Good work. Instead of using for loop, you can make your functions reduced to only 1 by suing `for in` loop.
which looks like `for(key in obj){...}`, key being the desk1....desk5, obj[key] being 'stu'...'sheldon'.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+      return this.charAt(0).toUpperCase() + this.slice(1);
+    }
+    
+    var nameToFind = "Fred";
+    
+    findStudent(nameToFind);
+    
+    function findStudent(studentName) {
+      console.log(studentName);
+      studentArray[findStudentIndex(studentName)] ? formatResults(studentArray[findStudentIndex(studentName)]) : failedResults(studentName);
```
joshuabyun: nice ternary!

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,151 @@
...
+    }
+    
+    function findStudentIndex(studentName) {
+      studentName = studentName.toTitleCase();
+      var counter = 0,
+          foundKid = false;
```
joshuabyun: when do you set this variable to true? isn't foundKid will always be false?

joshuabyun:
LF2

PR #83, opened 2017-01-12
js functions - Michael Duong

tallergeese:


js_functions/index.html
```diff
@@ -12,21 +12,25 @@
...
     <div class="card">
         <div class="front"><img src="imgs/path.jpg" alt=""></div>
-        <div class="back"><img src="imgs/card_design.png" alt="Card Back"></div>
+        <div class="back" onclick="flipCard(this)"><img src="imgs/card_design.png" alt="Card Back"></div>
     </div>
 </div>
 
+<button onclick="myMessage()">Wotcher!</button>
+<button onclick="add(5, 8)">ADD</button>
+<button onclick="console.log('Your sum is', add2result)">ADD #2</button>
```
collette-tamez: You labeled your console logs, NICE 😎

js_functions/main.js
```diff
@@ -0,0 +1,17 @@
+function myMessage(){
+    console.log("This is my first function!");
+}
+
+function add(x, y){
+    console.log("Your sum is", x + y);
```
collette-tamez: In this function and your `add2` it would be more efficient to save your result to a variable and then console log that value. 
```Javascript
function add(num1, num2){
     var result = num1 + num2;
     console.log("My function result is: " + result);
}
////OR
function add2(num1, num2){
     var secondResult = num1 + num2;
     return secondResult;
}
```

js_functions/main.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+function add2(x, y){
+    return x + y;
+}
+
+var add2result = add2(24, 72);
+
+function flipCard(element){
+    $(element).hide();
```

collette-tamez:
#LFZ2

PR #86, opened 2017-01-13
Js dom creation - John Yosas

jayosas75:


js_dom_creation/index.html
```diff
@@ -1,109 +1,136 @@
...
+    </style>
+
+    <script>
+        $(document).ready(function () {
+            //playground 1 code
+            var playground1_h1 = $('<h1>').html('I am a heading');
+            var playground1_p1 = $('<p>').html('This is an awesome paragraph');
+            var playground1_p2 = $('<p>').css('color', 'red').html('This has red text');
+
+            $('#playground1').append(playground1_h1, playground1_p1, playground1_p2);
```
collette-tamez: Nice work using your `append` here to append everything in a nice clean sweep.

js_dom_creation/index.html
```diff
@@ -1,109 +1,136 @@
...
+
+            //playground 2 code
+            var playground2_div1 = $('<div>').addClass('awesome_class').html('Here is some awesome div text');
+            var playground2_div2 = $('<div>').addClass('awesome_class').html('Here is MORE awesome div text');
+            var playground2_div3 = $('<div>').addClass('awesome_class').html('Here is <strong>EVEN MORE</strong> awesome div text');
+           /* did not add ID or 'data-coolness' to playground2_div3. If i add an ID to another element on the page it
+           wouldn't recognize it since its in the html already(i think). Also googled data-coolness, nothing came up :/*/
```
collette-tamez: Yeah (unfortunately?) `data-coolnes` isn't actually a thing. It is however a _custom attribute_, and there a certainly ways of adding custom attributes to your jQuery elements. For instance you could take the object approach to making your element.

js_dom_creation/index.html
```diff
@@ -1,109 +1,136 @@
...
+
+            $('#playground2').append(playground2_div1, playground2_div2, playground2_div3);
+
+            //playground 3 code
+            var playground3_ul = $('<ul>').addClass('neato');
+            var playground3_li1 = $('<li>').html('Link 1');
+            var playground3_li2 = $('<li>').html('Link 2');
+            var playground3_li3 = $('<li>').html('Link 3');
```
collette-tamez: I would suggest using a `for` loop here to make your `li` elements. That way you have block of code to make and append as many list items as you need.

PR #89, opened 2017-01-13
js dom creation - Charles Kim

ck1114:
In "playground destination 3," the addition of the inline <span> tag was messing up everything. I couldn't figure out a way to add it properly so I took it out.

### DIFF
--- a/js_dom_creation/index.html
+++ b/js_dom_creation/index.html
@@ -41,10 +41,41 @@
     <script>
         $(document).ready(function () {
             //playground 1 code
+            var h1 = $("<h1>").text("I am a heading");
+            var p1 = $("<p>").text("This is an awesome paragraph");
+            var p2 = $("<p>",{
+                text: "This has red text",
+                style: "color:red",
+            });
+            $(".playground_destination#playground1").append(h1, p1, p2);
 
             //playground 2 code
+            var pg2Div1 = $("<div>").addClass("awesome_class").text("Here is some awesome div text");
+            var pg2Div2 = $("<diV>").addClass("awesome_class").attr("id", "my_awesome_div2").text("Here is MORE awesome div text");
+            var pg2Div3 = $("<div>").addClass("awesome_class").attr("id", "my_awesome_div3").attr("data-coolness", "high").html("Here is <strong>EVEN MORE</strong> awesome div text");
+            $("#playground2.playground_destination").append(pg2Div1, pg2Div2, pg2Div3);
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo("#playground2.playground_destination");
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
+            //(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
 
             //playground 3 code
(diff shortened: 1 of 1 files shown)


js_dom_creation/index.html
```diff
@@ -41,10 +41,41 @@
...
+                text: "This has red text",
+                style: "color:red",
+            });
+            $(".playground_destination#playground1").append(h1, p1, p2);
 
             //playground 2 code
+            var pg2Div1 = $("<div>").addClass("awesome_class").text("Here is some awesome div text");
+            var pg2Div2 = $("<diV>").addClass("awesome_class").attr("id", "my_awesome_div2").text("Here is MORE awesome div text");
+            var pg2Div3 = $("<div>").addClass("awesome_class").attr("id", "my_awesome_div3").attr("data-coolness", "high").html("Here is <strong>EVEN MORE</strong> awesome div text");
+            $("#playground2.playground_destination").append(pg2Div1, pg2Div2, pg2Div3);
```
collette-tamez: Its good that you used append here to add everything at once.

js_dom_creation/index.html
```diff
@@ -41,10 +41,41 @@
...
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo("#playground2.playground_destination");
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
+            //$(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
+            //(pg2Div1, pg2Div2, pg2Div3).appendTo(".playground_destination#playground2");
```
collette-tamez: Moving forward you'll want to remove these big blocks of unused code. It will keep your files much neater and cleaner.

js_dom_creation/index.html
```diff
@@ -41,10 +41,41 @@
...
 
             //playground 3 code
+            var pg3_ul = $("<ul>").addClass("neato");
+            var pg3_1st_li = $("<li>").text("Link 1");
+            var pg3_2nd_li = $("<li>").text("Link 2");
+            var pg3_3rd_li = $("<li>").text("Link 3");
```
collette-tamez: Definitely try using a `for` loop here to create your list. That way you only have one block of code to create and append your elements. Let me know if you want to take some time to go through it together.

js_dom_creation/index.html
```diff
@@ -41,10 +41,41 @@
...
+            var pg3_figure = $("<figure>").addClass("awesome_picture");
+            //var pg3_img = $("<img>").attr("src", "http://comerecommended.com/wp-content/uploads/2013/01/how-to-get-awesome-back.jpg");
+            var pg3_img = $("<img>", {
+                src: "http://comerecommended.com/wp-content/uploads/2013/01/how-to-get-awesome-back.jpg"
+            });
+            //var pg3_figcaption = $("<figcaption>").html("That's right, <span class="bold_text">you are</span>");
```
collette-tamez: Now that I take a closer look at it, I believe that the reason this didn't work is because your quotation marks. Javascript doesn't care about which ones you use as long as you start and end with the same ones. The only time it does care is if you have nested quotes like you do here. So instead of using double quotes around your `class` you will want to approach it differently. 
```Javascript
var pg3_figcaption = $("<figcaption>").html('That\'s right, <span class="bold_text">you are</span>');
\\\\OR
var pg3_figcaption = $("<figcaption>").html("That's right, <span class='bold_text'>you are</span>");
```

PR #93, opened 2017-01-13
Js comparisons and conditionals - Ken Wolfson

kswolfson:


js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+        console.log("output1 = "+output1);
+        console.log("output2 = "+output2);
+        console.log("output3 = "+output3);
+        console.log("output4 = "+output4);
+
+        //Feature Set 3
+
+        var gameNumber = 64;
+        var userInput1 = 50;
+        var userInput2 = gameNumber;  //this way vs var named userInput2 = 64
```
collette-tamez: Excellent work setting up this variable 👍

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+
+        if (userInput1 > gameNumber)
```
collette-tamez: you really want to include the curly braces here to keep things organized. 
```Javascript
if ( userInput1 > gameNumber ){
     //code
} else {

///OR 

if (userInput1 > gameNumber)
{
     //code to be executed
}
else 
{
```

js_comparisons_conditionals/index.html
```diff
@@ -0,0 +1,74 @@
...
+            console.log("the number is too high");
+
+        if (userInput1 < gameNumber)
+            console.log("the number is too low");
```
collette-tamez: nice work using two if statements here since we wanted you to make both comparisons.

js_comparisons_conditionals/index.html
kswolfson: Is brackets always a best practice even when not required? In the past I typically left them out for readability reasons when not required. But maybe that's a dated practice.

collette-tamez: Dan might have a different opinion on that but I would suggest it for the purposes of this class. When you start sharing your code with everyone else in the cohort, it will be more familiar and easy to read for them and we don't dive to deeply on whether or not brackets and semi-colons are required.

PR #95, opened 2017-01-13
js_repeat_blocks - Darin Estrin

darin-estrin:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,73 @@
...
+
+      for (var i = numbers.length - 1; i >= 0; i--){
+        console.log('The current number is ' + numbers[i]);
+      }
+
+      for (var i = 0; i < numbers.length; i++){
+        if (numbers[i] % 2 !== 0) {
+          console.log(numbers[i] + ' is an odd number of the numbers array.');
+        }
+      }
```
joshuabyun: All 3 loops looking good, You have the correct start and the end! As a suggestion, try solving the same problem with using array methods, such as forEach, slice, reverse.. etc. The code can be written lot more simpler and concise - easy to understand.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,73 @@
...
+
+      var mixedArray = [4, 56, 11, 13, 97, 22, 7, 36, 19, 5, 8, 10];
+
+      for (var i = 0; i < mixedArray.length; i++){
+        if (mixedArray[i] % 2 !== 0) {
```
joshuabyun: very good use of modulo!

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,73 @@
...
+      var classRoom = {
+        desk1: 'Stu',
+        desk2: 'Peggy',
+        desk3: 'LoneStar',
+        desk4: 'George',
+        desk5: 'Sheldon'
+      };
+
+      for (desk in classRoom) {
+        console.log(classRoom[desk] + ' is at ' + desk);
```
joshuabyun: simple and nice for in loop.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+
+      // Feature Set 3
+      var studentArray = [{name: "Bob", grade: 87}, {name: "Fred", grade: 82}, {name: "George", grade: 93}, {name: "Robert", grade: 85}];
+
+      function findStudent(name){
+        var i = 0;
+        while (i < studentArray.length){
```
joshuabyun: Also try using a boolean value to start and exit the while loop in case you need to keep running the function even after you exit from the while loop.

joshuabyun:
LF2

PR #97, opened 2017-01-13
js dom creation - brock patterson

brockcp:
still trying to figure out git/gitthub. sorry for all the files attached here when it should just be the one on js dom creation.


js_dom_creation/index.html
```diff
@@ -40,11 +40,42 @@
...
     <script>
         $(document).ready(function () {
+
             //playground 1 code
+            var h1 = $('<h1>').text('I am a heading');
+            var p = $("<p>").text("This is an awesome paragraph");
+            var p_style_red = $("<p>").css({'color': 'red'}).text("This has red text");
+            $(h1).appendTo('#playground1');
+            $(p).appendTo("#playground1");
+            $(p_style_red).appendTo("#playground1");
```
collette-tamez: Make sure you understand the differences between `append` and `appendTo`. You could certainly use `append` to apply everything at once.

js_dom_creation/index.html
```diff
@@ -40,11 +40,42 @@
...
+            $(div1).appendTo("#playground2");
+            $(div2).appendTo("#playground2");
+            $(div3).appendTo("#playground2");
 
             //playground 3 code
+            $("ul").addClass("neato");
+            var ul_neato = $('<ul>').addClass("neato");
+            var neato_li1 = $('<li>').addClass('neato > li').text('link1');
+            var neato_li2 = $('<li>').addClass("neato > li").text('link2');
+            var neato_li3 = $('<li>').addClass("neato > li").text('link3');
```
collette-tamez: If I were you I would use a `for` loop here. That way you would only have to write one block of code to create and append as many list items as you need.

PR #98, opened 2017-01-13
Js dom creation - Michael Duong

tallergeese:


js_dom_creation/index.html
```diff
@@ -42,10 +42,48 @@
...
+                text:"Here is some awesome div text"
+            });
+            var p2_div2 = $("<div>", {
+                addClass:"awesome_class",
+                text:"Here is MORE awesome div text"
+            });
+            var p2_div3 = $("<div>", {
+                addClass:"awesome_class",
+                html:"Here is <strong>EVEN MORE</strong> awesome div text",
+                // I don't know how to add a data-coolness attribute with jquery, but I think awesome_class adds plenty of cool all by itself.
```
collette-tamez: Applying a custom attribute (or attributes in general) is exactly where the object method comes in handy. Try adding `data-coolness: "high"` to your object.

js_dom_creation/index.html
```diff
@@ -42,10 +42,48 @@
...
+            });
+
+            $("#playground2").append(p2_div1, p2_div2, p2_div3);
+
             //playground 3 code
+
+            var p3_nav = $("<ul>").addClass("neato");
```
collette-tamez: A good habit to get into would be using `$` for any an element selector stored variables like this. The dollar sign itself isn't really a special character so its okay to use. Also the dollar sign will give you an even clearer indication about what your variable is doing.
`var $p3_nav = $("<ul>").addClass("neato");`

js_dom_creation/index.html
```diff
@@ -42,10 +42,48 @@
...
+            var p3_fig = $("<figure>").addClass("awesome_picture");
+            var p3_figcap = $("<figcaption>").text("That's right, ");
+            var p3_figcap_bold = $("<span>").addClass("bold_text").text("you are");     //This span has the .bold_text class in the html, but the class isn't ever actually used for anything.
+            var p3_img = $("<img src='http://comerecommended.com/wp-content/uploads/2013/01/how-to-get-awesome-back.jpg'>");
+
+            for (var i=1; i<4; i++){
+                var p3_li = $("<li>").text("Link " + i);
+                p3_li.appendTo(p3_nav);
```
collette-tamez: Excellent work using a loop here 👏  👏

PR #99, opened 2017-01-14
js repeat blocks - Mark Crisostomo

mcrisostomo9:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,70 @@
...
+    <title>JS Repeat Blocks prototype</title>
+    <script>
+        //fs1
+        var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        var mixedArray = [4, 56, 11, 13, 97, 22, 7, 36, 19, 5, 8, 10];
+        for(var i = 0; i < numbers.length; i++){
+            console.log('Ascending array ' + numbers[i]);
+        }
+        console.log(' ');
+        for(var i = 9; i >= 0; i--){
```
joshuabyun: good set up of start and the end of the loop. 9 also could have been used from the number.length-1  similar to your first loop.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,70 @@
...
+            console.log('Descending array ' + numbers[i]);
+        }
+        console.log(' ');
+        for(var i = 0; i < numbers.length; i+=2){
+            console.log('Odds only numbers array ' + numbers[i]);
+        }
+        console.log(' ');
+        for(var i = 0; i < mixedArray.length; i++){
+            if(mixedArray[i] % 2 !=0){
```
joshuabyun: nice use of modulo!

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,70 @@
...
+        }
+        console.log(' ');
+        //fs2
+        var classRoom = {desk1: 'Stu',
+                         desk2: 'Peggy',
+                         desk3: 'LoneStar',
+                         desk4: 'George',
+                         desk5: 'Sheldon'}
+        for(var desk in classRoom){
+            console.log(classRoom[desk] +' is at ' + desk + '.');
```
joshuabyun: very nice use of key and value inside  a for in loop

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,70 @@
...
+            var i = 0;
+            var found = false;
+            while (!found && i < studentArray.length){
+                var student = studentArray[i];
+                if (student.name == name){
+                    output = 'Found ' + name + ' their grade is: ' + student.grade;
+                    found = true;
+                }
+                i++;
+                if (!found){
```
joshuabyun: since you are setting the value found in line 51, why not just define the value output right below?

joshuabyun:
LF2

PR #100, opened 2017-01-14
js repeating blocks - Michael Duong

tallergeese:
if I try to use findStudent() via the console, I get the right output, but it also throws up an undefined line. Not sure what's happening there.

### DIFF
--- a/js_repeat_blocks/index.html
+++ b/js_repeat_blocks/index.html
@@ -35,11 +35,37 @@
 
         var classRoom = {desk1: 'Stu', desk2: 'Peggy', desk3: 'LoneStar', desk4: 'George', desk5: 'Sheldon'};
 
-        var seatingString = "";
-
         for (var personIndex in classRoom){
             console.log(classRoom[personIndex] + " is at " + personIndex);
         }
+
+        //feature set 3
+
+        var studentArray = [{name: "Bob", grade: 87}, {name: "Fred", grade: 82}, {name: "George", grade: 93}, {name: "Robert", grade: 85}];     //i think there was a typo on the prototype instructions. originally the comma after Robert was a period
+
+        function findStudent(studentName){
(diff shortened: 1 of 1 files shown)


tallergeese:
The JS Dom Creation prototype instructions asked us to commit and push more frequently...

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,74 @@
...
+        //feature set 1
+
+        var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+
+        for (var i = 0; i <numbers.length; i++){
+            console.log("numbers array's values in ascending place in index: ", numbers[i]);
+        }
+
+        for (var i = numbers.length; i > 0; i--){
+            console.log("numbers array's values in descending place in index: ", numbers[i-1]);      //numbers array index goes from 0 to 9, but numbers.length starts counting at 1, so I need to subtract 1 from numbers.length to actually get the right place in the index. is this the right way to do it?
```
collette-tamez: So you could subtract here or you could subtract inside of the set up for your loop. 
`var i = numbers.length -1;` 
Its the same values but this way the subtraction only happens once when you declare the value of `i`. Then once you've made that change you can set your conditional to ` i >= 0` so it stops at zero.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        for (var i = 0; i < numbers.length; i++){
+            if (numbers[i]%2 != 0){     //modulus to check if there's a remainder when dividing by 2
```
collette-tamez: Nice work using modulus. While its tempting to base your loop on the index of the numbers you can't guarantee that you'll always have a neat ordered array to work with. Ideally you want your loops (and functions in general) to be able handle all types of random values.

PR #102, opened 2017-01-14
js repeat blocks - Gwenever Pacifico

gpacifico:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,48 @@
...
+        <title>Repeat Blocks</title>
+        <script>
+            var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+            for(var i = 0; i < numbers.length; i++) {
+                console.log(numbers[i]);
+            }
+            for(var i = numbers.length - 1; i >= 0; i--) {
+                console.log(numbers[i]);
+            }
+            for(var i = 0; i < numbers.length; i+=2) {
```

PR #103, opened 2017-01-15
js repeat blocks - Charles Kim

ck1114:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,59 @@
...
+<head>
+    <script>
+        //Start Feature Set 1 below
+        var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        for(var i=0; i<numbers.length; i++){
+            console.log(numbers[i]);
+        }
+
+        //following is the 'for loop' for the numbers 10-1 in descending order
+        for(var i=9; i<=numbers.length; i--){
```
collette-tamez: Be very careful when looping backwards! You are very close with this one but you `<=` is causing an infinite loop. If `i` becomes a negative number is it still less than or equal to `number.length`?

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,59 @@
...
+            console.log(numbers[i]);
+        }
+        console.log("**following is the output of odd numbers from 1-10**")
+        for(var i=0; i<numbers.length; i++){
+            console.log(numbers[i++]);
```

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,59 @@
...
+        }
+        var mixedArray = [4, 56, 11, 13, 97, 22, 7, 36, 19, 5, 8, 10];
+        for(var i=0; i<mixedArray.length; i++) {
+            console.log(mixedArray[i++]);
+        }
+        //the for loop above doesn't output only the odd numbers, so
+        for(var i=0; i<mixedArray.length; i++) {
+            if(mixedArray[i]%2 == 1){
```
collette-tamez: Oops i almost didn't see your other loop here. I was about to scold your for looping through mixed array improperly. 😉

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,59 @@
...
+                console.log(mixedArray[i]);
+            }
+        }
+
+        //Start Feature Set 2 below
+        var classRoom = {desk1: 'Stu', desk2: 'Peggy', desk3: 'LoneStar', desk4: 'George', desk5: 'Sheldon'}
+        for(var desk in classRoom){
```
collette-tamez: Your `for in` loop looks correct. Personally I like `for in` because its much easier to visualize.

PR #104, opened 2017-01-15
js dom creation - Yaeri Lim

yaerilim:


js_dom_creation/index.html
```diff
@@ -41,10 +41,26 @@
     <script>
         $(document).ready(function () {
             //playground 1 code
-
+                var playground1_h1 = $("<h1>").text('I am a heading');
+                var playground1_p1 = $("<p>").text('This is an awesome paragraph');
+                var playground1_p2 = $("<p>").css('color', 'red').text('This has red text');
```
collette-tamez: Very nice descriptive names here 👍

js_dom_creation/index.html
```diff
@@ -41,10 +41,26 @@
...
+                $('#playground1').append(playground1_h1, playground1_p1, playground1_p2);
             //playground 2 code
-
+                var playground2_div1 = $("<div>").addClass('awesome_class').text('Here is some awesome div text');
+                var playground2_div2 = $("<div>").addClass('awesome_class').attr('id', 'my_awesome_div2').text('Here is MORE awesome div text');
+                var playground2_div3 = $("<div>").addClass('awesome_class').attr('id', 'my_awesome_div3').css('data-coolness', 'high').html('Here is <strong>EVEN MORE</strong> awesome div text');
```
collette-tamez: `data-coolness` is whats called a "custom attribute", which means its basically made up. But since its an attribute you would be better of using `.attr()` to apply it. This is also where using the object method of dom creation comes in handy.

js_dom_creation/index.html
```diff
@@ -41,10 +41,26 @@
...
+                $('#playground2').append(playground2_div1, playground2_div2, playground2_div3);
             //playground 3 code
+                var playground3_ul = $('<ul>').addClass('neato');
+                var playground3_li1 = $("<li>").text('Link 1');
+                var playground3_li2 = $("<li>").text('Link 2');
+                var playground3_li3 = $("<li>").text('Link 3');
```
collette-tamez: Try using a `for` loop here to create your list items. That way you only have to write one block of code to create and append your elements.

PR #106, opened 2017-01-15
js primitive data types - Ken Wolfson

kswolfson:


js_primitive_data_types/index.html
```diff
@@ -0,0 +1,36 @@
...
+        var output1 = "First Name: "+firstName+",  Last Name: "+lastName+",  Age: "+age;
+        console.log(output1);
+
+        //Feature Set 3
+
+        var x = 25;
+        var y = 75;
+
+        var output2 = x+y;
+        output3 = x.toString()+" and "+y.toString();
```
collette-tamez: If you are already concatenating a string with your variables you don't need to use `.toString()` here. That method would be more effectively used if you were only using variables like in your `output2`. So for instance if you wanted your output2 to be a string rather than a number you could apply the `.toString()` method to it.

PR #107, opened 2017-01-15
js functions - Ken Wolfson

kswolfson:


js_functions/main.js
```diff
@@ -0,0 +1,19 @@
+function myMessage(){
+    console.log("It's Alive!!");
+}
+
+function add(x, y){
+    var z=x+y;
+    console.log("x+y="+z);
```
collette-tamez: Its a best practice to assign your result to a variable like you did here. I would suggest implementing in your `add2` function.

js_functions/main.js
```diff
@@ -0,0 +1,19 @@
...
+function add2(x, y){
+    return x+y;
+}
+
+function cardFlip(card){
+    $(card).hide();
+}
+
+var add2result = add2(10, 36);
+var recursiveResult = add2(add2(5, 10), add2(20, 30));
```
collette-tamez: Nice work on this one here.

PR #109, opened 2017-01-15
js spec data types - Ken Wolfson

kswolfson:


js_special_data_types/index.html
```diff
@@ -0,0 +1,47 @@
...
+
+        var output4 = typeof(x);
+        var output5 = typeof(y);
+        var output6 = typeof(z);
+
+        //I also typed the following in the console per prototype instructions
+
+        console.log("typeof output4 is "+output4);
+        console.log("typeof output5 is "+output5);
+        console.log("typeof output6 is "+output6);
```
collette-tamez: Assigning your output to a variable and then logging that variable rather than placing your logic inside of your console is great work.

PR #112, opened 2017-01-16
Js dom creation Jorge Cruz

RobzDoom:


js_dom_creation/index.html
```diff
@@ -36,15 +36,58 @@
             font-style: italic;
             font-size: 120%;
         }
+        .bold_text{
+            text-align: center;
+            font-weight: bold;
+            font-style: italic;
+            font-size: 120%;
```
collette-tamez: You've got the right thinking here by using CSS, however you can instead apply this CSS directly using jQuery.

js_dom_creation/index.html
```diff
@@ -36,15 +36,58 @@
...
+        }
+
     </style>
 
     <script>
         $(document).ready(function () {
             //playground 1 code
+            var the_h1 = $('<h1>').text('I am heading');
+            var p_1 =  $('<p>').text('This is an awesome paragraph');
+            var p_2 =  $('<p>').css("color", "red").text('This is red text');
```
collette-tamez: Be mindful of your names, try not to abbreviate too much. It would be better to have clear and concise names even if they're long winded.

js_dom_creation/index.html
```diff
@@ -36,15 +36,58 @@
...
+
+            $(the_h1).appendTo('#playground1');
+            $(p_1).appendTo('#playground1');
+            $(p_2).appendTo('#playground1');
```
collette-tamez: Be sure to not the differences between `append` an `appendTo`. You can actually append all your elements to `#playground1` all at once if you use `.append()`

js_dom_creation/index.html
```diff
@@ -36,15 +36,58 @@
...
+
+            $(divOne).appendTo('#playground2');
+            $(divTwo).appendTo('#playground2');
+            $(divThree).appendTo('#playground2');
 
             //playground 3 code
+            var aUl = $('<ul>').addClass('neato');
+            var li1 = $('<li>').text('Link 1');
+            var li2 = $('<li>').text('Link 2');
+            var li3 = $('<li>').text('Link 3');
```
collette-tamez: Try using a loop to generate your list items. That way you only have to write one block of code to generate and append your your `li`. Being able to dynamically generate elements using a loop is skill that will definitely come in handy.

PR #113, opened 2017-01-16
Js repeat blocks

RobzDoom:


js_repeat_blocks/index.html
```diff
@@ -0,0 +1,50 @@
...
+    <meta charset="UTF-8">
+    <title>js_Repeat_Blocks</title>
+        <script>
+            var numbers = [1,2,3,4,5,6,7,8,9,10];
+            var mixedArray = [4, 56, 11, 13, 97, 22, 7, 36, 19, 5, 8, 10];
+            var classRoom = {desk1: 'Stu', desk2: 'Peggy', desk3: 'LoneStar', desk4: 'George', desk5: 'Sheldon'};
+            var studentArray = [{name: "Bob", grade: 87}, {name: "Fred", grade: 82}, {name: "George", grade: 93}, {name: "Robert", grade: 85}];
+
+            for (var i = 0; i <= numbers.length; i++){
+                    console.log(i);
```
collette-tamez: You should definitely label your console logs, especially when you just have three loops in a row logging numbers like this.

js_repeat_blocks/index.html
```diff
@@ -0,0 +1,50 @@
...
+                }
+            for (var i = numbers.length; i >= 0; i--){
+                    console.log(i);
+                }
+
+            for (var i = 1; i <= numbers.length; i+=2){
+                    console.log(i)
```

PR #114, opened 2017-01-16
js ref data types - Derrick Sebesta

dsebesta:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,57 @@
...
+        student.lastName = 'Sebesta';
+        student.age = '34';
+        student.school = 'LearningFuze';
+        student.questionList = [];
+        student.hasQuestion = false;
+        student.greeting = function() {
+            return `Hello, my name is ${this.firstName}`;
+        };
+        student.aboutMe = function() {
+            return `My full name is ${this.firstName} ${this.lastName}, I am ${this.age} years old and I am a student at ${this.school}.`;
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,57 @@
...
+        };
+        student.askQuestion = function(question) {
+            this.hasQuestion = true;
+            this.questionList.push(question);
```
collette-tamez: Very nice work here.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,57 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>JS Ref Data Types</title>
+    <script>
+        var students = ['Derrick', 'Ken', 'Michael', 'John', 'Gwen'];
+        console.log(students[2]);
+        console.log(students);
+        console.log(students.length);
+        students.push('Darin', 'Daniel');
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,57 @@
...
+            if (this.hasQuestion) {
+                this.hasQuestion = false;
+                return this.questionList[this.questionList.length - 1];
+            }
+            else {
+                return 'You have no pending questions';
+            }
+        };
+
+        console.log('Greeting function:', student.greeting());
```
collette-tamez: Beautiful work labeling your console logs. 👏

PR #115, opened 2017-01-16
js_ref_data_types - Darin Estrin

darin-estrin:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>js-ref-data-types</title>
+  </head>
+  <script>
+
+    // Feature Set 1
+    console.log('Feature Set 1');
```
collette-tamez: Beautiful work here using comments and console logs to organize your file. I would suggest that you start removing empty lines from your files. That way everything is nice and compact.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,72 @@
...
+    var student_array = ['Darin', 'John', 'Dan', 'Mike', 'Matt'];
+    console.log('The third student is ' + student_array[2]);
+
+    // Feature Set 1.1
+    console.log('The complete student list is ' + student_array);
+    console.log('The length of the student_array is ' + student_array.length);
+    student_array.push('Steffan', 'Amy');
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,72 @@
...
+    student.firstName = 'Darin';
+    student.lastName = 'Estrin';
+    student.age = 30;
+    student.school = 'Learningfuze';
+    student.questionList = [];
+    student.hasQuestion = false;
+
+    // Feature Set 2.1
+    student.greeting = function(){
+      return 'Hello, my name is ' + this.firstName;
```
collette-tamez: Very nice work here. I would also suggest assigning your string a variable. It would be more efficient that way to simply return of console log that variable rather than the entire string.

PR #116, opened 2017-01-16
js ref data types - Daniel Fogerty

DanFog:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <meta charset="UTF-8">
+    <title>JS Reference Data Types</title>
+
+    <script src="https://code.jquery.com/jquery-2.1.4.js"></script>
+    <script>
+        var student_array = ["Daniel Fogerty", "Andy Ong", "Derrick Sebesta", "Gwenever Pacifico", "Ken Wolfson"];
+        console.log(student_array[2]);
+        console.log(student_array);
+        console.log(student_array.length);
+        student_array.push("Yaeri Lim", "Tim Feng");
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+        //feature set 2
+        var student = {};
+        student.firstName = "Daniel";
+        student.lastName = "Fogerty";
+        student.age = 22;
+        student.school = "LearningFuze";
+        student.questionList = [];
+        student.hasQuestion = false;
+        student.greeting = function() {
+            return "Hello my name is " + student.firstName;
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+        student.aboutMe = function() {
+          return "My full name is " + student.firstName + ", I am " + student.age + " years old and I am a student at " + student.school + ".";
+        };
+        student.askQuestion= function(question) {
+            student.questionList.push(question);
+            student.hasQuestion = true;
+            return "A question has been added to your list";
+        };
+        student.answerQuestion = function() {
+            if (student.hasQuestion == false) {
```
collette-tamez: Nice work on your conditional here. In general its more efficient to use `===` over `==`, of course there are use cases for both but technically the triple equals is faster.

PR #117, opened 2017-01-16
Js ref data types - Michael Duong

tallergeese:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>JS Reference Data Types</title>
+    <script>
+        var student_array = ["John", "Ken", "Michael", "Tim", "Charles"];
+        console.log("The third student is", student_array[3]);
```
collette-tamez: Remember that arrays start at index 0. The first item in your array would be `student_array[0]`, it has an _index of zero but its still the first item in your array_. So`student_array[3]` is pointing to the fourth item in your array.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,60 @@
...
+        console.log(student_array);
+        console.log(student_array.length);
+
+        student_array.push("Gwenever", "Andy");
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,60 @@
...
+
+        student.firstName = 'Michael';
+        student.lastName = 'Duong';
+        student.age = 26;
+        student.school = 'LearningFuze';
+        student.questionList = [];
+        student.hasQuestion = false;
+
+        student.greeting = function() {
+            return "Hello, my name is " + student.firstName;
```

PR #119, opened 2017-01-16
js ref data types - John Yosas

jayosas75:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,70 @@
...
+    <title>JS Reference Data Types</title>
+    <script>
+        //feature set 1.0
+        var student_array = ['John', 'Gwen', 'Derrick', 'Brock', 'Dan'];
+        console.log(student_array[2]);
+
+        //feature set 1.1
+        console.log(student_array);
+        console.log(student_array.length);
+        student_array.push('Andy', 'Mark');
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,70 @@
...
+        student.school = 'LearningFuze';
+        student.questionList = [];
+        student.hasQuestion = false;
+
+        //feature set 2.1
+        student.greeting = function() {
+            return 'Hello, my name is ' + this.firstName;
+        }
+        student.aboutMe = function() {
+            return 'My full name is ' + this.firstName + " " + this.lastName + ", I am " + this.age +
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,70 @@
...
+                    ' years old and I am a student at ' + this.school + '.';
+        }
+        student.askQuestion = function(question) {
+            this.hasQuestion = true;
+            this.questionList.push(question);
+            return 'A question has been added to your list'
+        }
+
+        student.answerQuestion = function() {
+            if (this.hasQuestion = true) {
```
collette-tamez: Make sure you note the differences between `=` `==` and `===`.

With your use of `=` all you're doing here is giving your `student.hasQuesion` a value of `true`.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,70 @@
...
+                return this.questionList[this.questionList.length - 1];
+            } else {
+                return 'You have no pending questions';
+            }
+        }
+
+        //feature set 2.2
+        console.log('My greeting: ' + student.greeting());
+        console.log('About me: ' + student.aboutMe());
+        console.log('answerQuestion Function: ' + student.answerQuestion());
```
collette-tamez: Look very closely at what this console log is returning. Does it return the value you've assigned to your `else` statement? Again, revaluate your conditional in this function.

PR #120, opened 2017-01-16
Js ref data types Jorge Cruz

RobzDoom:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>js_ref_data_types</title>
+    <script>
+        var studentArray = ['Nelson', 'Lisa', 'Bart', 'Millhouse', 'Ralph'];
+            console.log(studentArray[2]); //Should be Bart
+            console.log(studentArray.length);
+        studentArray.push('Martin', 'Todd');
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+        student.age = 10;
+        student.school = 'Springfield Middle School';
+        student.questionList = [];
+        student.hasQuestion = false;
+        console.log(student);
+        student.greeting = function(){
+            return("Hello, my name is "+ student.firstName);
+        };
+        student.aboutMe = function (){
+          return 'My full name is ' + student.firstName +' '+ student.lastName + ', I am' +' '+ student.age + ' years old and I am a student at' +' '+ student.school + '.';
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+        };
+        student.askQuestion = function (question) {
+            student.questionList.push(question);
+            student.hasQuestion = true;
+            return "A question has been added to your list";
+        };
+        student.answerQuestion = function () {
+            if (student.hasQuestion === true){
```

PR #121, opened 2017-01-16
js ref data types - Andy Ong

Breathics:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,66 @@
...
+    <title>JS Ref Data Types</title>
+    <script src="https://ajax.aspnetcdn.com/ajax/jQuery/jquery-3.1.1.min.js"></script>
+    <script>
+        // Feature Set 1.0 - Student Array
+        var student_array = ['Charles', 'Jorge', 'Andy', 'Yaeri', 'John']; // Names suspended in an array
+        console.log(student_array[2]);
+
+        // Feature Set 1.1 - Array Methods
+        console.log(student_array);
+        student_array.length;
```
collette-tamez: `.length` here isn't really doing anything. Length is less of a method and more of a value. Think of the object you constructed right below this. If you were to `console.log(student.firstName)` you would get back the value `"Andy"`. In this case `.length` would act the exact same way, so you would need to console log it to see the value.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,66 @@
...
+        student_array.push('Mark');
+        student_array.push('Darin');
```

You can also add more than one value using one `push`.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,66 @@
...
+        student.questionList = [];
+        student.hasQuestion = false;
+
+        // Feature Set 2.1 - Add methods to the madness
+        student.greeting = function(){
+            return 'Hello my name is ' + student.firstName;
+        };
+
+        student.aboutMe = function() {
+            return 'My full name is ' + student.firstName + ' ' + student.lastName + ', I am ' + student.age +
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,66 @@
...
+        student.askQuestion = function(question) {
+            student.questionList.push(question);
+            student.hasQuestion = true;
+            return 'A question has been added to your list';
+        };
+
+        student.answerQuestion = function() {
+            if (student.hasQuestion == true){
+                student.hasQuestion = false;
+                return student.questionList.pop();
```
collette-tamez: `.pop()` will return the last item in your array, however we wanted you to preserve your array structure. How would you return the last item without removing it from the array?

PR #122, opened 2017-01-16
js ref data types - Charles Kim

ck1114:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,64 @@
...
+<html>
+<head>
+    <script>
+        //Start Feature Set 1.0
+        var student_array = ["Charles", "Naree", "Alyssa", "Brandon", "Eddy"];
+        console.log(student_array[2]);
+        //Start Feature Set 1.1
+        console.log(student_array);
+        console.log(student_array.length);
+        student_array.push("Jay", "Mark");
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,64 @@
...
+        student.age = 32;
+        student.school = "LearningFuze";
+        student.questionList = [];
+        student.hasQuestion = false;
+        //Start Feature Set 2.1
+        student.greeting = function(){
+            return "Hello, my name is " + student.firstName;
+        }
+        student.aboutMe = function(){
+            return "My full name is " + student.firstName +", I am " + student.age + " years old and I am a student at " + student.school+".";
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,64 @@
...
+        }
+        student.askQuestion = function(question){
+            student.questionList.push(question);
+            student.hasQuestion = true;
+            return "A question has been added to your list."
+        }
+        student.answerQuestion = function(){
+            if(student.hasQuestion === true){
+                return student.questionList[student.questionList.length-1];
```

PR #123, opened 2017-01-16
js ref data types - Tim Feng

spikers:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,73 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <script>
+    //Feature Set 1.0
+    var student_array = ["Tim", "Michael", "George", "Charles", "Gwen"];
+    console.log(student_array[3-1]);
```
collette-tamez: Thats kind of an obtuse way of doing it but I mean it works.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,73 @@
...
+
+    //Feature Set 1.1
+    console.log(student_array);
+    console.log(student_array.length);
+    student_array.push("Joe", "Zorah");
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,73 @@
...
+    console.log(student_array);
+
+    //Feature Set 2.0
+    var student = {};
+    student.firstName = "Tim";
+    student.lastName = "Feng";
+    student.age = "105";
```
collette-tamez: Old man Tim!!!

js_ref_data_types/index.html
```diff
@@ -0,0 +1,73 @@
...
+    student.questionList = [];
+    student.hasQuestion = false;
+
+    //Feature Set 2.1
+    student.greeting = function() {
+        return "Hello, my first name is " + this.firstName;
+    };
+    
+    student.aboutMe = function() {
+        return "My full name is " + this.firstName + " " + this.lastName + ", I am " + this.age + " years old and I am a student at " + this.school + ".";
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,73 @@
...
+    };
+
+    student.askQuestion = function(question) {
+        this.questionList.push(question);
+        this.hasQuestion = true;
+        return "A question has been added to your list!";
+    };
+
+    student.answerQuestion = function() {
+        if (this.hasQuestion) {
```
collette-tamez: Excellent work on this part 👏  👏

js_ref_data_types/index.html
spikers: Super obtuse! I was worried that people would be unable to figure out where I got the [2] from haha.

PR #124, opened 2017-01-16
js ref data types - Gwenever Pacifico

gpacifico:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Reference Data Types</title>
+        <script>
+            var student_array = ["Maddy", "Janiece", "William", "Christine", "Henry"];
+            console.log("Third Student:", student_array[2]);
+            console.log("Full Array:", student_array);
+            console.log("Length:", student_array.length);
+            student_array.push("Jack", "Daniel");
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+            student.age = 32;
+            student.school = "Stargate Institute";
+            student.questionList = [];
+            student.hasQuestion = false;
+
+            student.greeting = function () {
+                return "Hello, my name is " + this.firstName;
+            };
+            student.aboutMe = function () {
+                return "My full name is " + this.firstName + " " + this.lastName + ", I am " + this.age + " years old and I am a student at " + student.school + ".";
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,58 @@
...
+            };
+            student.askQuestion = function (question) {
+                student.questionList.push(question);
+                student.hasQuestion = true;
+                return "A question has been added to your list.";
+            };
+            student.answerQuestion = function () {
+                if(student.hasQuestion === true) {
```

PR #125, opened 2017-01-16
js ref data types - Ken Wolfson

kswolfson:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,65 @@
...
+<title>JS Reference Data Types Prototype</title>
+
+<script>
+
+    var student_array = ["Ken", "Alycia", "David", "Tally", "Dashiell"];
+
+    console.log("3rd student: "+student_array[2]);
+
+    student_array.push("Quinn"); //add Quinn to student_array
+    student_array.push("Adrian"); //add Adrian to student_array
```
collette-tamez: With the push method you can add more than one value at a time.
`student_array.push("Quinn", "Adrian");`

There are quite a few methods for adding and removing things from an array. There are also quite a few array methods in general. Be sure to check them out, they come in handy

js_ref_data_types/index.html
```diff
@@ -0,0 +1,65 @@
...
+
+    var student_object = {
+
+        firstName:"Ken",
+        lastName: "Wolfson",
+        age: 20,
+        school: "BHS",
+        questionList: [],
+        hasQuestion: false
+    };
```
collette-tamez: You want to use the dot notation method for constructing your object. This is your first dabble into object oriented javascript. Its a very valuable skill.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,65 @@
...
+
+    function greeting () {
+        return "Hello, my name is " + student_object.firstName;
+    };
+
+    function aboutMe () {
+        return "My full name is " + student_object.firstName+" "+student_object.lastName+", " +
```
collette-tamez: I would suggest setting your string to variable, that way it would be more efficient just to return that variable rather than the whole string.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    function askQuestion (question) {
+        student_object.questionList.push(question);
+        student_object.hasQuestion = true;
+        return "A question has been added to your list";
+    };
+
+    function answerQuestion () {
+        if (student_object.hasQuestion) {
```
collette-tamez: Nice work on this part 👍

PR #126, opened 2017-01-16
js ref data types - Yaeri Lim

yaerilim:


js_ref_data_types/index.html
```diff
@@ -0,0 +1,61 @@
...
+    <title>Title</title>
+    <script>
+        //Feature Set 1.0 - Student Array
+        var student_array = ['Alan', 'Bob', 'Chad', 'David', 'Eddy'];
+        console.log(student_array[2]);
+
+        //Feature Set 1.1 - Array Methods
+        console.log(student_array);
+        console.log(student_array.length);
+        student_array.push('Fred', 'Greg');
```

js_ref_data_types/index.html
```diff
@@ -0,0 +1,61 @@
...
+        student.school = "Learningfuze";
+        student.questionList = [];
+        student.hasQuestion = false;
+
+        //Feature Set 2.1 - Add methods to the madness
+        student.greeting = function () {
+            return "Hello, my name is " + this.firstName;
+        };
+        student.aboutMe = function () {
+            return "My full name is " + this.firstName + " " + this.lastName + ", I am " + this.age + " years old and I am a student at " + this.school + ".";
```
collette-tamez: Good work on these ones. I would suggest setting your string to variable, that way it would be more efficient just to return that variable rather than the whole string.

js_ref_data_types/index.html
```diff
@@ -0,0 +1,61 @@
...
+        };
+        student.askQuestion = function(question) {
+            this.questionList.push(question);
+            this.hasQuestion = true;
+            return "A question has been added to your list";
+        };
+        student.answerQuestion = function () {
+            if (this.hasQuestion == true) {
+                this.hasQuestion = false;
+                return this.questionList.slice(-1)[0];
```
collette-tamez: For this part we wanted you to return the last item in the array but also leave the array intact. `splice` can come in handy but really all you have to do is return the index of the last item in the array.

PR #131, opened 2017-01-20
Js event delegation - Charles Kim

ck1114:


js_event_delegation/script.js
```diff
@@ -1,10 +1,25 @@
...
 // Answer the questions here:
 
 // - What do you think is going to happen?
-
+// It will display the text "Button #5"
 // - What happened?
-
+// The text wasn't displayed
 // - Why?
+// I don't know!!!!
```
collette-tamez: Because of event delegation (or lack thereof)! So the code the executes in a document ready happens from top to bottom. So on line 2, a click handler gets applied to all of the buttons nested inside `#list` that jQuery can find. Then on line 5 your new button gets generated. The button didn't exist when your click handlers got applied, just like how your click handlers won't be applied if they aren't inside a document ready because the dom hasn't loaded yet. 

When you use the `.on` to delegate the event the click handler applies to all of the buttons regardless of when a button gets generated. Instead of looking at the individual buttons, we're instead looking at the parent element `$("#list")` and attaching the event handler to `#list` and in our `.on("click", "button",` we're telling our event to stop and execute when it reaches a `button`

PR #134, opened 2017-01-20
Js event delegation - Mark Crisostomo

mcrisostomo9:


js_event_delegation/script.js
```diff
@@ -1,10 +1,28 @@
...
 
 // - What do you think is going to happen?
-
+// Delegated button will console.log the name of the button.
 // - What happened?
-
+//The button was correctly logged to the console.
 // - Why?
-
+//Because the click delegated the function to the newly created button
```
collette-tamez: Pretty much, rather than look at the individual buttons we instead look at the `list` element and let the event bubble until it reaches the target button 👍  Dynamically generated elements are the perfect use case for event delegation with `.on`

js_event_delegation/script.js
```diff
@@ -1,10 +1,28 @@
...
 //========== Write your code below ===========//
-
+$(document).ready(function () {
+    //fs1
+    $('#list').on('click','button', function () {
+        console.log($(this).text() + " was clicked.");
```
collette-tamez: Nice work using `this` and labeling your console log 👏  👏

js_event_delegation/script.js
```diff
@@ -1,10 +1,28 @@
...
+    });
+    //fs2
+    var new_button = $("<li><button style='margin-top: 10px'>Delegated Button#5 Handler</button></li>");
+    $('#list').append(new_button);
+    //fs3
+    $('#list').on('click', 'li:last-child button', function () {
+       console.log('Newest button that was added was clicked.');
+    })
+    //additional challenge
+    var google_button = $("<li><button class='google' style='margin-top: 10px'><a style='text-decoration: none' target='_blank' href='https:www.google.com'> Go to Google</a></button></li>");
```
collette-tamez: Also check out the `window.target` method

http://stackoverflow.com/questions/19851782/how-to-open-a-url-in-a-new-tab-using-javascript-or-jquery

js_event_delegation/script.js
```diff
@@ -1,10 +1,28 @@
...
+    $('#list').append(google_button);
+    $('#list').on('click', 'button[class=google]', function(){
```
collette-tamez: Nice work on your selectors. Definitely try using a conditional to accomplish the same thing on your original function on line 2.

PR #135, opened 2017-01-20
js_event_delegation - Michael Duong

tallergeese:


js_event_delegation/script.js
```diff
@@ -2,9 +2,64 @@
...
 
+//The new element will not activate the click handler.
+
 // - What happened?
 
+//The new element did not activate the click handler
+
 // - Why?
 
+//The new button did not exist at the time the event handlers were called.
```
collette-tamez: Exactly 👍  Keep event delegation with dynamically generated elements in mind as it will definitely come in handy with your future projects.

js_event_delegation/script.js
```diff
@@ -2,9 +2,64 @@
...
+    });
+
+    var newLi = $('<li>');
+
+    $(newLi).append(newButton);
+    $("#list").append(newLi);
+
+//Feature Set 3
+
+    $('#list').on('click','button',function(){
```
collette-tamez: Be sure you note the difference between this function and your feature set 1 function and how the consoles change if you were to toggle the two functions. Also the order of your functions matter in a document ready.

js_event_delegation/script.js
```diff
@@ -2,9 +2,64 @@
...
+    $("#list").append(newerLi);
+
+    $('#list').on('click','button', function(event){
+
+        var coolCheck = event.target;
+        console.log('coolCheck is ' + coolCheck);
+        var attr = $(coolCheck).attr('datacoolness');
+        console.log('attribute: ' + attr);
+        if (attr !== undefined){
+            window.open("http://www.google.com");
```
collette-tamez: Be sure to include the `_blank` as a second parameter to open in a new tab.

PR #136, opened 2017-01-20
js event delegation - Andy Ong

Breathics:


js_event_delegation/script.js
```diff
@@ -1,10 +1,33 @@
...
+// The .on method is not applied to the new button
+// It is appended after the .on method is called to '#list'
+// There will be no console.log
 
 // - What happened?
+// We were also able to get the delegated button to log out as well
 
 // - Why?
+// The .on method is applied to list rather than a new button
+// We are appending onto #list when #list still has the .on method
```
collette-tamez: Spot on 👍  Be sure to keep event delegation with dynamically generated elements in mind as it will definitely be needed for future projects.

js_event_delegation/script.js
```diff
@@ -1,10 +1,33 @@
...
+$(document).ready(function(){
+
+    $('#list button').on('click', function(){
+        console.log("Button :", event);
+        console.log("Button :", this);
+    });
+
+    $('#list').append($('<li><button style="margin 10px">Delegated Button#5 Handler</button></li>'));
+
+    $('#list').on('click', function(){
```
collette-tamez: remember that you have to pass a second parameter of the element you want to target. Currently your click handler is only applied to your `ul`. 
screen shot 2017-01-20 at 12 51 32 pm

But this is still a good example of event delegation. Even though you click on a button the function still execute because your buttons are nested inside of your `#list`. The event fires anytime you click anywhere on the element.

PR #137, opened 2017-01-20
js event delegation - Gwenever Pacifico

gpacifico:


js_event_delegation/script.js
```diff
@@ -1,10 +1,35 @@
...
+    $("#list button").on("click", function () {
+        console.log($(this).text());
+    });
+
+    //Feature Set 2
+    var button = $("<li><button style='color: red'>Child Button 5</button></li>");
+    $("#list").append(button);
+
+    //Feature Set 3
+    $("#list").on("click", "button", function () {
```
collette-tamez: Be sure you note the difference between this function and your feature set 1 function and how the consoles change if you were to toggle the two functions. Also the order of your functions matter in a document ready.

js_event_delegation/script.js
```diff
@@ -1,10 +1,35 @@
...
+        console.log($(this).text());
+    });
 
+    //Additional Challenge
+    var newButton = $("<li><button data-cheese = 'yellow'><a href='https://www.google.com/' target='_blank'>Cheese Button</a></button></li>");
```
collette-tamez: Also check out the `window.target` method

http://stackoverflow.com/questions/19851782/how-to-open-a-url-in-a-new-tab-using-javascript-or-jquery

js_event_delegation/script.js
```diff
@@ -1,10 +1,35 @@
...
+    $("#list").append(newButton);
+    $("#list").on("click", "button", function () {
+        if($(event.target).attr("data-cheese") !== undefined) {
```
collette-tamez: Nice work on this part. Try adding this conditional block to your above function.

PR #138, opened 2017-01-20
Js event delegation - Yaeri Lim

yaerilim:


js_event_delegation/script.js
```diff
@@ -1,10 +1,29 @@
...
 
 // - What do you think is going to happen?
-
+//nothing
 // - What happened?
-
+//nothing
 // - Why?
-
+//Becuase the 5th button was made after console.log($(this).text());???
```
collette-tamez: Kind of. Its because the button gets created after the click handler is applied with ` $('#list button').on('click', function () {`. Inside of a document ready the order matters, and your code runs top to bottom. 
1. Jquery looks for every button inside `#list` the applies the click handler.
2. Your new button gets created. 
Because step 1 already happened you button won't do anything when you clicked on it. Instead you want to delegate your event to `#list`

js_event_delegation/script.js
```diff
@@ -1,10 +1,29 @@
...
 //========== Write your code below ===========//
 
+$(document).ready(function() {
+    $('#list button').on('click', function () {
+        console.log($(this).text());
+    });
+
+    $('#list').append("<li><button style='margin-top: 10px'>Delegated Button#5 Handler</button></li>");
+
+    $('#list button').on('click', function () {
```
collette-tamez: Try rewriting this function with by using event delegation. Instead of targeting `#list button` target only list, then pass in `button` as a second parameter. If you add `button` to your function you are telling your click handler to look for any button inside of `#list` regardless of when it gets created. 

http://www.w3schools.com/jquery/event_on.asp

js_event_delegation/script.js
```diff
@@ -1,10 +1,29 @@
...
+        console.log($(this).text());
+    });
+
+    $('#list').append("<li><button id='google'>Open Google!</button></li>");
+
+    $('#list').on('click', '#google', function() {
+        if($(event.target).attr('id') !== undefined){
```
collette-tamez: This conditional will work but not in your current function, try moving it into one of your above functions. It won't work as intended because by targeting the `id` you will always get the button, your `if` statement will always be true, and therefore unnecessary.

PR #145, opened 2017-01-24
js oop - Mark Crisostomo

mcrisostomo9:


js_oop/index.html
```diff
@@ -0,0 +1,63 @@
...
+        function coin_stack(initial_amount) {
+            this.value = initial_amount;
+            this.add_coins = function(added_coins) {
+                this.value += added_coins;
+            }
+            this.remove_coins = function(subtracted_coins) {
+                if (this.value >= subtracted_coins){
+                    this.value -= subtracted_coins;
+                    return subtracted_coins;
+                }else if(this.value < subtracted_coins){
```
collette-tamez: Very nice work on this function. Your use of `-=` is excellent and your use of an `else if` here is also great 👍

PR #147, opened 2017-01-24
js oop - Andy Ong

Breathics:


js_oop/index.html
```diff
@@ -0,0 +1,58 @@
...
+            this.value = initial_coins;
+//      Feature Set 2
+            this.add_coins = function(inc_coins){
+                this.value += inc_coins;
+            };
+            this.remove_coins = function(dec_coins){
+                if (this.value >= dec_coins){
+                    this.value -= dec_coins;
+                    return dec_coins
+                } else if (this.value < dec_coins){
```
collette-tamez: Excellent work on this function. Your use of `-=` and your `else if` are spot on. Although I would suggest going with either a simpler parameter name or I would recommend not abbreviating them.  As it is I find `inc_coins` and `dec_coins` to be somewhat vague. Even it feels long something like `coinsToAdd` is much more straight forward.

Breathics: Sounds good, that's probably the thing that stops me sometimes. Banana is true.

PR #149, opened 2017-01-24
js oop - Gwenever Pacifico

gpacifico:


js_oop/index.html
```diff
@@ -0,0 +1,56 @@
...
+                this.add_coins = function(coin_amount_to_be_added) {
+                    this.value += coin_amount_to_be_added;
+                };
+                this.remove_coins = function(coin_amount_to_be_removed) {
+                    if(this.value >= coin_amount_to_be_removed) {
+                        this.value -= coin_amount_to_be_removed;
+                        return coin_amount_to_be_removed;
+                    }
+                    else {
+                        coin_amount_to_be_removed = this.value;
```
collette-tamez: Very nice work on this function here. You've got an excellent naming convention 👍

js_oop/index.html
```diff
@@ -0,0 +1,56 @@
...
+            console.log("I have " + my_coin_stack.value + " coins in my coin stack");
+            my_coin_stack.remove_coins(37);
+            console.log("I have " + my_coin_stack.value + " coins in my coin stack");
+            my_coin_stack.remove_coins(55);
+            console.log("I have " + my_coin_stack.value + " coins in my coin stack");
+
+            coin_stack.prototype.transfer_coins = function(object_coins_transferred_from, amount_of_coins_being_transferred) {
+                var new_amount_to_remove = object_coins_transferred_from.remove_coins(amount_of_coins_being_transferred);
+                this.add_coins(new_amount_to_remove);
+                return amount_of_coins_being_transferred;
```
collette-tamez: Look back to your `remove_coins` function. You included an `else` block to account for when the amount you were removing was more than what was available in the stack. So the amount that gets actually transferred may not be the amount that was requested. Rather than returning your parameter you would be better off returning `new_amount_to_remove` since that would have a more accurate amount.

PR #151, opened 2017-01-31
js_ajax - Michael Duong

tallergeese:


js_ajax/main.js
```diff
@@ -9,8 +10,21 @@ $(document).ready(function(){
             url: 'http://ax.itunes.apple.com/WebObjects/MZStoreServices.woa/ws/RSS/topMovies/json',
             success: function(result) {
                 console.log('AJAX Success function called, with the following result:', result);
-
+                global_result = result;
+                first_movie = result.feed.entry[0]["im:image"][2].label;
+                for (var i = 0; i < result.feed.entry.length; i++){
+                    var movie_image = result.feed.entry[i]["im:image"][2].label;
+                    //the directors are in the title property for some reason too, but I'll keep both for completeness, although it's a bit redundant
+                    var movie_info = "Title: " + result.feed.entry[i].title.label + " Director: " + result.feed.entry[i]["im:artist"].label;
```
collette-tamez: You could've also used the `result.feed.entry[i]["im:name"].label` which is also in each entries object and has the title of the film only.

PR #152, opened 2017-01-31
js ajax - Tim Feng

spikers:


js_ajax/main.js
```diff
@@ -9,9 +18,45 @@ $(document).ready(function(){
...
             success: function(result) {
                 console.log('AJAX Success function called, with the following result:', result);
+                //Feature Set 1
+                global_result = result;
+
+
+                //Feature Set 2
+                first_movie_image = global_result.feed.entry[0]["im:image"][2].label;
+
+                first_movie_info = global_result.feed.entry[0];
```
collette-tamez: Notice how you're using `.feed.entry` a lot. You could make your life easier by assigning `global_result = result.feed.entry`. That way you only have to access `global_results[0]`. Also If I were you I would have switched these two variables around and left `first_movie_info` as an object. 
`first_movie_info = global_result[0]`
`first_movie_image = first_movie_info["im:image"][2].label`
Nice and clean and only capturing what you need.

js_ajax/main.js
```diff
@@ -9,9 +18,45 @@ $(document).ready(function(){
...
 
+                //Feature Set 3
+                for (var i = 0; i < global_result.feed.entry.length; i++) {
+                    movie_images.push(global_result.feed.entry[i]["im:image"][2].label);
+                    //Feature Set 4
+                    movie_directors.push(global_result.feed.entry[i]["im:artist"].label)
+                    movie_titles.push(result.feed.entry[i]["im:name"].label);
+                }
+                itemAppending(movie_images, movie_directors, movie_titles);
```
collette-tamez: Clever use of a function. Keep this exact structure you just implemented in this prototype in mind as you'll need to be referencing it soon (_hint hint_)

PR #157, opened 2017-02-01
js ajax - Andy Ong

Breathics:


js_ajax/main.js
```diff
@@ -1,15 +1,41 @@
...
+            method: 'get',
             url: 'http://ax.itunes.apple.com/WebObjects/MZStoreServices.woa/ws/RSS/topMovies/json',
             success: function(result) {
                 console.log('AJAX Success function called, with the following result:', result);
+                // Feature Set 1
+                global_result = result;
+                console.log(global_result);
 
+                // Feature Set 2
+                var first_movie = result.feed.entry["0"];
```
collette-tamez: Here you were meant to have stored the third image of the first movie. When accessing `.entry` its an array so there no need to put the index in quotes. Quotes are only need when you need to access certain keys in the object (like `im:name`). You've also already saved you `result` object to your `global_result` why not use it? Better yet make your life easier by assigning `global_result = result.feed.entry` that way you can later on access the array of films `global_result[0]`

js_ajax/main.js
```diff
@@ -1,15 +1,41 @@
...
+
+                // Global Result variable set to hold 3rd image of the first movie
+                global_result = global_result.feed.entry["0"]["im:image"][2].label;
+                console.log(global_result);
+
+                // Feature Set 3
+                var movies = [];
+                var titleInfo = [];
+                for (var i = 0; i < result.feed.entry.length; i++){
+                    movies[i] = result.feed.entry[i]["im:image"][2].label;
```
collette-tamez: Interesting decision to assign the value to the index directly. You could've just have easily used push though theres nothing wrong with this method. The difference comes down to execution speed, apparently push is faster on chrome (at least).
https://jsperf.com/array-push-vs-unshift-vs-direct-assignment/2

js_ajax/main.js
```diff
@@ -1,15 +1,41 @@
...
+                    var image = $("<img>").attr('src', movies[i]);
+                    $('#main').append(image);
+
+                // Feature Set 4
+                    titleInfo[i] = result.feed.entry[i].title.label;
+                    var title = $("<p>").text(titleInfo[i]);
+                    $("#main").append(title);
```
collette-tamez: It feels a bit unnecessary to push everything to an array if you're still appending in your loop. Why not skip the middle man and access your `global_results` directly? If anything you would be better of pushing your elements to an array and then appending them all at once outside of your loop.

PR #158, opened 2017-02-05
js ajax - Mark Crisostomo

mcrisostomo9:


js_ajax/main.js
```diff
@@ -9,7 +9,15 @@ $(document).ready(function(){
             url: 'http://ax.itunes.apple.com/WebObjects/MZStoreServices.woa/ws/RSS/topMovies/json',
             success: function(result) {
                 console.log('AJAX Success function called, with the following result:', result);
-
+                global_result = result.feed.entry;
```
collette-tamez: Nice work saving your saving your response down to the `entry`  object.

js_ajax/main.js
```diff
@@ -9,7 +9,15 @@ $(document).ready(function(){
...
+                for(i=0;i < global_result.length-1; i++){
```
collette-tamez: So if `global_result.length` is `10` and `global_result.length-1` is `9` but you're only looping until `i` is _less than_ `9` then you'll only get `global_result[8]`. This only gets you the first nine results instead of all ten.

PR #159, opened 2017-02-06
js ajax - Charles Kim

ck1114:


js_ajax/main.js
```diff
@@ -7,9 +8,23 @@ $(document).ready(function(){
...
             dataType: 'json',
             url: 'http://ax.itunes.apple.com/WebObjects/MZStoreServices.woa/ws/RSS/topMovies/json',
-            success: function(result) {
+            success: function (result) {
                 console.log('AJAX Success function called, with the following result:', result);
-
+                global_result = result;
+                var first_movie_info = global_result.feed.entry[0]["im:image"][2].label;
+                console.log("the third image from the first movie", first_movie_info)
+                for(var i=0; i<global_result.feed.entry.length; i++) {
```
collette-tamez: Notice how you're using `.feed.entry` a lot. You could make your life a little easier by assigning `global_result = result.feed.entry` that way you can make use of `global_result[0]`. Its simple and clean but that way you're only storing and utilizing the part of the response object you need.

js_ajax/main.js
```diff
@@ -7,9 +8,23 @@ $(document).ready(function(){
...
+                    var images = $("<img>", {src: global_result.feed.entry[i]["im:image"][2].label});
+                    var figcaption = $("<figcaption>");
+                    var movie_title = $("<h3>",{
+                        text: global_result.feed.entry[i]["im:name"].label
+                    });
+                    var movie_director = $("<h4>", {
+                        text: global_result.feed.entry[i]["im:artist"].label
+                    });
+                    figcaption.append(movie_title, movie_director);
```
collette-tamez: http://www.w3schools.com/tags/tag_figcaption.asp

`<figcaption>` needs to be the last child in a `<figure>` element. If you're unfamiliar with an element always make sure to double check that you're using it correctly.

PR #160, opened 2017-02-06
Js ajax - Yaeri Lim

yaerilim:


js_ajax/main.js
```diff
@@ -10,6 +12,29 @@ $(document).ready(function(){
             success: function(result) {
                 console.log('AJAX Success function called, with the following result:', result);
 
+                //Feature Set 1 - Road to success
+                global_result = result;
+                console.log('global_result: ' + global_result);
```
collette-tamez: Look at the difference between whats printed in the console from 13 vs whats printed in the console from line 17.
screen shot 2017-02-06 at 11 13 07 am

Be careful when combining strings the use of your `+` here means your console log is trying to print your `global_result` as a string. You would be better off using a comma like in line 13, that way you can still see your whole object.

js_ajax/main.js
```diff
@@ -10,6 +12,29 @@ $(document).ready(function(){
...
+
+                //Feature Set 2 - Exploring for images
+                var first_movie_info = result.feed.entry[0];
```
collette-tamez: You've already saved your `result` to your `global_result` variable, you should use it. Better yet save yourself some typing by assigning `global_results = result.feed.entry` that way you only have to use `global_result[0]`

js_ajax/main.js
```diff
@@ -10,6 +12,29 @@ $(document).ready(function(){
...
+                console.log('first_movie_info: ' + first_movie_info);
+                var first_movie_img = result.feed.entry[0]['im:image'];
+                console.log('first_movie_img: ' + first_movie_img);
+                var first_movie_third_img = result.feed.entry[0]['im:image'][2];
```
collette-tamez: You're on the right track but here you're still printing out the entire object. You want to instead print the image source located inside of the image.

js_ajax/main.js
```diff
@@ -10,6 +12,29 @@ $(document).ready(function(){
...
+                console.log('first_movie_third_img: ' + first_movie_third_img);
+
+                //Feature Set 3 - Adding the Images & Feature Set 4 -Title the Images
+                var third_img_from_movies = [];
+                var movie_title_director = [];
```
collette-tamez: You have these variables defined as arrays but you're not utilizing them as arrays. You would be better off defining them like `global_result` is on line 4 `var global_result;`

PR #161, opened 2017-02-06
Js ajax Jorge Cruz

RobzDoom:


js_ajax/main.js
```diff
@@ -9,9 +8,32 @@ $(document).ready(function(){
...
+                console.log(arrivalMovieThirdImage);
+
+                var threeImages = [];
+                var moviesInformation = [];
+                for (var i = 0; i <global_result.feed.entry.length; i++){ ///Loop through movies.
+                    threeImages[i] = global_result.feed.entry[i]['im:image'][2].label; // Find the THIRD img from each movie.
+                    var images = $('<img>').attr('src', threeImages[i]); //Make var create image with src and threeImages[i]
+                    $('#main').append(images); //Append var images to id main in HTML
 
+                    moviesInformation[i] = global_result.feed.entry[i].title.label;
```
collette-tamez: Interesting decision to assign the value to the index directly. You could've just have easily used push thought theres nothing wrong with this method. The difference comes down to execution speed, apparently push is faster on chrome (at least).
https://jsperf.com/array-push-vs-unshift-vs-direct-assignment/2

PR #163, opened 2017-02-15
php_includes - Daniel Fogerty

DanFog:
Done and Doner


php_includes/index.php
```diff
@@ -0,0 +1,16 @@
+<?php
+  $a = 1;
+  $b = 5;
+  include('includes/data.php');
+  $result = $a * $b;
+  print("<br>Result is $result");
+?>
+
+<?php
```
collette-tamez: Unless you're adding in html or something like that theres no need to break up your tags like this.

php_includes/index.php
```diff
@@ -0,0 +1,16 @@
...
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
```

PR #164, opened 2017-02-15
PHP Intro - Tim Feng

spikers:


php_intro/to.js
```diff
@@ -1,35 +1,108 @@
...
+var my_int = parseInt(my_float);
+var my_float2 = 5.3;
+var my_int2 = parseInt(my_float2);
+
+//EXERCISE 6
+
+var my_var1 = 'Hello';
+var my_var2;
 
+if(my_var2){
```
collette-tamez: Keep in mind that with `isset()`, php checks that the variable exists (aka not undefinied) and also checks if the variable is _not_ null.

PR #166, opened 2017-02-15
Php intro Jorge Cruz

RobzDoom:


php_intro/to.js
```diff
@@ -1,35 +1,103 @@
...
+var my_float2 = 5.3;
+var my_int2 = parseInt(my_float2);
 
+//EXERCISE 6
 
+var my_var1 = 'Hello';
 
-//EXERCISE 7
-
+if(my_var2){
```
collette-tamez: Keep in mind that with `isset()`, php checks that the variable exists (aka not undefined) and also checks if the variable is _not_ null. So to be correct about it, you would still need to define `var my_var2;`. If you look in the console this throws an error in Javascript because the variable doesn't exist at all.

php_intro/to.js
```diff
@@ -1,35 +1,103 @@
...
+student = {};
+student.name = 'Skippy';
+student.class = 'English';
+student.grade = 75;
 
 //EXERCISE 10
+var num_array = [35, 2, 14, 56, 65, 52];
+
+    function find_the_greatest_num_and_index(direction) {
+        num_array;
```
collette-tamez: Remember that global variables operate differently in php. We use ` global $num_array;` not because we're defining the variable but because the variable is already defined but exists outside the scope of our function. The way you've written this here its not really doing anything and surprisingly your function still runs.

php_intro/to.js
```diff
@@ -1,35 +1,103 @@
...
+            }
+            i += increment;
+        }
+        return { 'greatest': greatest,
+                 'greatest_index': greatest_index
+        };
+
+    }
+
+find_greatest_num_and_index(1);
```
collette-tamez: You named your function something different :(

PR #167, opened 2017-02-15
PHP Includes - Tim Feng

spikers:


php_includes/index.php
```diff
@@ -0,0 +1,18 @@
...
+  $b = 5;
+  include('includes/data.php');
+  $result = $a * $b;
+  print("<br>Result is $result");
+?>
+
+<?php
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
```

php_includes/includes/main.php
```diff
@@ -0,0 +1,3 @@
+<main id="main_content">
+  Main Content
```
collette-tamez: You didn't add an image 🤒

PR #168, opened 2017-02-15
php_intro - Daniel Fogerty

DanFog:


php_intro/to.js
```diff
@@ -1,35 +1,114 @@
...
+var my_int = parseInt(my_float);
+var my_float2 = 5.3
+var my_int2 = parseInt(my_float2);
 
+//EXERCISE 6
 
-//EXERCISE 7
+var my_var1 = 'Hello';
 
+if(typeof my_var2 !== 'undefined') {
```

I find that the best resource for php is their official documentation (just avoid the user submitted notes)

php_intro/to.js
```diff
@@ -1,35 +1,114 @@
...
+    this.class = studentClass;
+  };
+}
+
+student = new studentConstructor;
+console.log(student);
+student.setName('Skippy');
+student.setGrade(75);
+student.setClass('English')
+console.log(student);
```
collette-tamez: Nice work being so thorough 🥇

PR #169, opened 2017-02-15
Php intro - Yaeri Lim

yaerilim:


php_intro/to.js
```diff
@@ -1,35 +1,106 @@
...
+var my_float = 3.5;
+var my_int = parseInt(my_float);
+var my_float2 = 5.3;
+var my_int2 = parseInt(my_float2);
+
+//EXERCISE 6
 
+var my_var1 = 'Hello';
+
+if(my_var2 !== null){
```
collette-tamez: Keep in mind that with `isset()`, php checks that the variable exists (aka not undefined) and also checks if the variable is _not_ null. So to be correct about it, you would still need to define `var my_var2;` or look at the `typeof` for this variable. If you look in the console this throws an error in Javascript because the variable doesn't exist at all.

php_intro/to.js
```diff
@@ -1,35 +1,106 @@
...
+var student = {
+    name: 'Skippy',
+    class: 'English',
+    grade: 75
+};
 
 //EXERCISE 9
 
-
+var student = new stdClass;
```
collette-tamez: Definitely go back and review OOP. You're on the right track with this but before you can use a _new_ class you have to define it first. This also throws an error in your console.

PR #170, opened 2017-02-15
php intro - michael duong

tallergeese:


php_intro/to.js
```diff
@@ -1,35 +1,111 @@
...
-
-
+var $variable1 = 'hello';
+var $variable2 = 4;
+var $variable3 = ['foo', 'bar'];
 
 //EXERCISE 2
-
-
+$variable1 = 4;
```
collette-tamez: I mean using the dollar sign is certainly thorough but its arguable whether its accurate. Some like to implement a dollar sign in their javascript variables to denote dom elements (kind of like how jquery does).

php_intro/to.js
```diff
@@ -1,35 +1,111 @@
...
+var $my_float = 3.5;
+var $my_int = Math.floor($my_float);
+var $my_float2 = 5.3;
+var $my_int2 = Math.floor($my_float2);
 
+//EXERCISE 6
+
+var $my_var1 = 'Hello';
+
+if(typeof $my_var2 !== 'undefined'){
```

php_intro/to.js
```diff
@@ -1,35 +1,111 @@
...
-
+var student = new studentObj('Skippy');
+console.log(student);
 
+//EXERCISE 10
+var $num_array = [35, 2, 14, 56, 65, 52];
+
+function find_greatest_number_and_index(direction){
+
+    var $greatest = $num_array[0];
```
collette-tamez: Interesting choice. I guess I would feel better about this choice if the array we were looking at gets passed in a parameter, that way your function is more versatile.

php_intro/to.js
```diff
@@ -1,35 +1,111 @@
...
+        incrementer = -1;
+    }
+    while (i != endLoop){
+        if ($num_array[i] > $greatest) {
+            $greatest = $num_array[i];
+            $greatest_index = i;
+        }
+        i = i + incrementer;
+    }
+    return 'the greatest number is ' + $greatest + ' and the greatest index is ' + $greatest_index;
```
collette-tamez: I would've rather you return an object here.

PR #172, opened 2017-02-15
PHP intro - Mark Crisostomo

mcrisostomo9:


php_intro/to.js
```diff
@@ -1,35 +1,108 @@
...
+var my_float = 3.5;
+var my_int = parseInt(my_float);
+var my_float2 = 5.3;
+var my_int = parseInt(my_float2);
 
+//EXERCISE 6
+
+var my_var1 = 'Hello';
+
+if(my_var2){
```
collette-tamez: Keep in mind that with `isset()`, php checks that the variable exists (aka not undefined) and also checks if the variable is _not_ null.

PR #173, opened 2017-02-15
PHP intro - Charles Kim

ck1114:


php_intro/to.js
```diff
@@ -1,35 +1,97 @@
...
+}
+console.log("Result = " + result);
 
 //EXERCISE 4
-
+console.log("This is a string "+" This is another string");
 
 
 //EXERCISE 5
+var my_float = 3.5;
```
collette-tamez: You forgot to declare your float variable here

php_intro/to.js
```diff
@@ -1,35 +1,97 @@
...
+var my_int = parseInt(my_float);
+var my_int2 = parseInt(my_float);
 
+//Exercise 6
+var my_var1 = "Hello";
+
+if (my_var2){
```
collette-tamez: Keep in mind that with `isset()`, php checks that the variable exists (aka not undefined) and also checks if the variable is _not_ null.

php_intro/to.js
```diff
@@ -1,35 +1,97 @@
...
 
 
 
 //EXERCISE 10
 
-
+var num_array = [35, 2, 14, 56, 65, 52];
+
+function find_greatest_num_and_index(direction){
+    num_array;
```
collette-tamez: Remember that global variables operate differently in php. We use ` global $num_array;` not because we're defining the variable but because the variable is already defined but exists outside the scope of our function. The way you've written this here its not really doing anything and surprisingly your function still runs.

PR #174, opened 2017-02-15
Php includes - Darin Estrin

darin-estrin:


php_includes/index.php
```diff
@@ -0,0 +1,18 @@
+<?php
+// Feature Set 1
+  $a = 1;
+  $b = 5;
+  include('includes/data.php');
+  $result = $a * $b;
+  print("<br>Result is $result");
+?>
+
+<?php
```
collette-tamez: Unless you're including html theres really no need to break up your php like this.

php_includes/index.php
```diff
@@ -0,0 +1,18 @@
...
+  //Feature Set 2
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
+  include('includes/ data2.php');
+  $result2 = $c;
+  print("<br>Result 2 = $result");
```
collette-tamez: You printed wrong variable here. 🤒

php_includes/index.php
```diff
@@ -0,0 +1,18 @@
...
+  include('includes/data.php');
+  $result = $a * $b;
+  print("<br>Result is $result");
+?>
+
+<?php
+  //Feature Set 2
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
```

PR #175, opened 2017-02-15
Php intro - Darin Estrin

darin-estrin:


php_intro/to.js
```diff
@@ -1,35 +1,105 @@
...
 
 
 //EXERCISE 3
+var numbers = [3, 5, 16, 3, 4, 18];
+var result = numbers[0];
 
+for (var i = 0; i < numbers.length; i++) {
+  result += numbers[i];
+}
+console.log(numbers);
```
collette-tamez: You should be labeling your console logs

php_intro/to.js
```diff
@@ -1,35 +1,105 @@
...
 
 //EXERCISE 5
+var my_float = 3.5;
+var my_int = parseInt(my_float);
+var my_float2 = 5.3;
+var my_int2 = parseInt(my_float2);
 
+// EXERCISE 6
+var my_var1 = "Hello";
+if (typeof my_var2 !== undefined) {
```

php_intro/to.js
```diff
@@ -1,35 +1,105 @@
...
+student.class = 'English';
+student.grade = 75;
 
 
 //EXERCISE 10
-
-
+var num_array = [35, 2, 14, 56, 65, 52];
+
+function find_greatest_num_and_index(arr, direction){
```
collette-tamez: 🥇  Very nice work making this function versatile rather than hardcoding it...

php_intro/to.js
```diff
@@ -1,35 +1,105 @@
...
+    }
+    i += increment;
+  }
+  console.log({
+    'greatest' : greatest,
+    'greatest_index': greatest_index
+  });
+}
+
+find_greatest_num_and_index(1);
```
collette-tamez: ...but you didn't pass it in when you made the function call 😢

PR #177, opened 2017-02-15
Php includes - Yaeri Lim

yaerilim:


php_includes/index.php
```diff
@@ -0,0 +1,15 @@
+<?php
+$a = 1;
+$b = 5;
+include('includes/data.php');
+$result = $a * $b;
+print("<br>Result is $result");
+?>
+<?php
```
collette-tamez: Unless you're including html theres really no reason to break up your php tags like this

php_includes/index.php
```diff
@@ -0,0 +1,15 @@
...
+$c = 2;
+include('includes/data2.php');
+include_once('includes/data2.php');
```

PR #181, opened 2017-02-16
PHP Includes - Charles Kim

ck1114:


php_includes/index.php
```diff
@@ -0,0 +1,16 @@
...
+$b = 5;
+include('includes/data.php');
+$result = $a * $b;
+print("<br>Result is $result");
+?>
+
+<?php
+$c = 2;
+include('includes/data2.php');
+include_once('includes/data2.php');
```

php_includes/index.php
```diff
@@ -0,0 +1,16 @@
+<?php
+$a = 1;
+$b = 5;
+include('includes/data.php');
+$result = $a * $b;
+print("<br>Result is $result");
+?>
+
+<?php
```
collette-tamez: Unless you're including html theres really no need to break your tags up like this.

PR #182, opened 2017-02-16
PHP includes - Mark Crisostomo

mcrisostomo9:


php_includes/index.php
```diff
@@ -0,0 +1,14 @@
+<?php
+  $a = 1;
+  $b = 5;
+  include ('includes/data.php');
+  $result = $a * $b;
+  print("<br />Result is $result");
+
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
```

php_includes/main.php
```diff
@@ -0,0 +1 @@
+This is my main ishhhh.
```
collette-tamez: You forgot to link an image 🤒

PR #184, opened 2017-02-16
PHP Includes - Andy Ong

Breathics:


php_includes/index.php
```diff
@@ -0,0 +1,14 @@
+<?php
+  $a = 1;
+  $b = 5;
+  require_once('includes/data.php');
```
collette-tamez: You've included these files but they're empty (at least on git). What happened here?

Breathics: @collette-tamez Pushed data.php and data2.php. Thanks for catching that!

php_includes/index.php
```diff
@@ -0,0 +1,14 @@
...
+  $result = $a * $b;
+  print("<br>Result is $result");
+
+  $c = 2;
+  include('includes/data2.php');
+  include_once('includes/data2.php');
```

PR #185, opened 2017-02-16
PHP Includes - Gwenever Pacifico

gpacifico:


php_includes/index.php
```diff
@@ -0,0 +1,17 @@
+<?php
+$a = 1;
+$b = 5;
+include('includes/data.php');
+$result = $a * $b;
+print("<br>Result is $result");
+?>
+
+<?php
```
collette-tamez: Theres no need to break up your tags like this unless you're including html

php_includes/index.php
```diff
@@ -0,0 +1,17 @@
...
+$c = 2;
+include('includes/data2.php');
+include_once('includes/data2.php');
```

PR #186, opened 2017-02-16
Php super globals - Yaeri Lim

yaerilim:


php_superglobals/index.php
```diff
@@ -0,0 +1,4 @@
+<?php
+print_r($_GET);
+var_dump($_POST);
```
collette-tamez: Make sure you note the differences between all the different print methods in php. They definitely have certain advantages.

php_superglobals/ajax.html
```diff
@@ -0,0 +1,48 @@
...
+    array(2) {
+    ["value"]=>
+    string(5) "howdy"
+    ["value2"]=>
+    string(5) "hello"
+    }-->
+    <script>
+        function ajax_call() {
+            $.ajax({
+                url: 'http://localhost/prototypes_C1.17/php_superglobals/index.php',
```
collette-tamez: since index.php exists within the same file structure all you need to do is point to this file. This URL won't resolve on my end.

PR #187, opened 2017-02-16
PHP Super Globals - Gwenever Pacifico

gpacifico:


php_superglobals/ajax.html
```diff
@@ -0,0 +1,32 @@
...
+                        success: function(response){
+                            console.log("Ajax call mission completed", response);
+                        }
+                    })
+                });
+            });
+        </script>
+    </head>
+    <body>
+        <form action="index.php">
```
collette-tamez: Form action automatically sends the data on submit. 
https://www.w3schools.com/tags/att_form_action.asp

You already have an ajax call so theres no need for the action here.

PR #192, opened 2017-02-16
PHP Super Globals - Charles Kim

ck1114:


php_superglobals/ajax.html
```diff
@@ -0,0 +1,30 @@
...
+    <script>
+        function ajaxCall(){
+            $.ajax({
+                dataType: "json",
+                data: {
+                    checkbox: $(".checkbox").val(),
+                    textarea: $(".textarea").val()
+                },
+                method: "post",
+                url: "http://localhost/prototypes_C1.17/php_superglobals/index.php",
```
collette-tamez: since index.php exists within the same file structure all you need to do is point to this file. This URL won't resolve on my end.

PR #193, opened 2017-02-16
Php super globals - Mark Crisostomo

mcrisostomo9:


php_superglobals/ajax.html
```diff
@@ -0,0 +1,40 @@
...
+                        console.log(result);
+                    }
+                });
+            })
+        })
+    </script>
+</head>
+
+<body>
+    <form class=""  method="post">
```
collette-tamez: Form action automatically sends the data on submit. 
https://www.w3schools.com/tags/att_form_action.asp

You already have an ajax call so theres no need for the action here.

PR #194, opened 2017-02-16
Php super globals

RobzDoom:


php_superglobals/ajax.html
```diff
@@ -0,0 +1,38 @@
...
+                            console.log('You are doing AJAX wrong!')
+                        }
+                    })
+                });
+            });
+    </script>
+</head>
+
+<body>
+    <form method="post">
```
collette-tamez: Since you're making an ajax request you don't need to specify any attributes here. Also you're setting two different methods so be careful.

php_superglobals/form.html
```diff
@@ -0,0 +1,45 @@
...
+                          console.log('You are doing AJAX wrong!')
+                      }
+                 })
+            });
+        });
+    </script>
+</head>
+<body>
+
+    <form action = "index.php" method="GET">
```
collette-tamez: Here we're trying to illustrate the difference between form submission and ajax. Please use one or the other, as you're asking for trouble with both.

PR #198, opened 2017-02-18
ang_dom_and_events - Darin Estrin

darin-estrin:


ang_dom_and_events/index.html
```diff
@@ -4,9 +4,19 @@
...
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.7/angular.min.js"></script>
+
 
 </head>
-<body>
+<body ng-app>
+
+  <div ng-init="iAmTrue=true; counter=0">
+    <span ng-hide="iAmTrue">The Button is Enabled</span>
+    <span ng-show="iAmTrue">The Button is Disabled</span>
```
collette-tamez: Be careful with your naming conventions. `iAmTrue` is a bit vague and at some point becomes false which of course effects `ng-hide`/`ng-show`. I would've gone super generic with something like `myBool`

PR #207, opened 2017-02-19
Ang dom events

RobzDoom:


ang_dom_and_events/index.html
```diff
@@ -2,11 +2,20 @@
...
     <meta charset="UTF-8">
-    <title>Angular DOM and Events</title>
+    <title>Angular Intro</title>
     <!-- Add Angular CDN below -->
-
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.7/angular.min.js"></script>
 </head>
-<body>
+<body ng-app>
+<div ng-init="trueVar = true" ng-init="counter = 0">
```
collette-tamez: Much like how you can't have multiple of the same the same attribute on an element (ie `<div class="foo" class="bar">`) you can't have multiple `ng-init`. You can initialize more than one variable separated by a semi-colon. `<div ng-init="trueVar = true; counter = 0">` 

Since your variable insn't initialized correctly it actually doesn't get created until you make the first click. Hence why there's no 0 value in your `he button has been clicked `

PR #215, opened 2017-02-20
ang_dom_events - Daniel Fogerty

DanFog:


ang_dom_and_events/index.html
```diff
@@ -4,9 +4,24 @@
...
     <!-- Add Angular CDN below -->
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.7/angular.min.js"></script>
 
 </head>
-<body>
-
+<body ng-app>
+  <div ng-init="var1='true'" ng-init="counter=0">
+    <span ng-show="var1">
+      Button is disabled
```
collette-tamez: Notice how your button starts disabled on page load even though the checkbox isn't clicked. 
screen shot 2017-02-20 at 2 57 24 pm

On line 11 you defined `var1` as equal to the _string_ `'true'` and not the boolean. It doesn't become an actual true or false value until you actually start toggling the checkbox.

PR #230, opened 2017-02-21
Angular Dom and Events - John Yosas

jayosas75:


ang_dom_and_events/index.html
```diff
@@ -4,9 +4,23 @@
...
+    <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.7/angular.min.js"></script>
 
 </head>
-<body>
-
-</body>
-</html>
\ No newline at end of file
+<body ng-app>
+<div ng-init="myVar=true" ng-init="counter=0">
```
collette-tamez: Much like how you can't have multiple of the same the same attribute on an element (ie `<div class="foo" class="bar">`) you can't have multiple `ng-init`. You can initialize more than one variable separated by a semi-colon. `<div ng-init="trueVar = true; counter = 0">` 

Since your variable isn't initialized correctly it actually doesn't get created until you make the first click. Hence why there's no 0 value in your `the button has been clicked `

PR #252, opened 2017-02-22
Mysql basics - Yaeri Lim

yaerilim:


mysql_basics/mysql_queries.sql
```diff
@@ -1 +1,17 @@
...
+--Select all todo_items made by user_id 123
+SELECT * FROM `items` WHERE `user_id` LIKE '123'
+--Insert a new item by user_id 123
+INSERT INTO `items` SET `title`='do laundry', `details`='hand wash blouses', `timestamp`=now(), `user_id`='123'
+--Delete a item by user_id 123
+DELETE FROM `items` WHERE `title`='homework'
+--Update a item by user_id 123
+UPDATE `items` SET `details`='hand wash dresses' WHERE `user_id`='123'
+--Perform a select to get all information from user_id 123
+SELECT * FROM `items` WHERE `user_id` LIKE '123'
```
TimelikeClosure: This should be getting information from the `` `users` `` table, not the `` `items` `` table.
> - Perform a select to get all information from `users` by using the user's id

mysql_basics/mysql_queries.sql
```diff
@@ -1 +1,17 @@
...
+INSERT INTO `students` SET `name`='New Student', `email`='imnew@email.com', `age`='35', `SSN`='456789123', `username`='addedstudent', `password`=sha1('123newbee'), `joined`=now(), `status`= 'active'
+DELETE FROM `students` WHERE `username`='addedstudent'
+
+--Feature Set 2 - Make a Table
+--Select all todo_items made by user_id 123
+SELECT * FROM `items` WHERE `user_id` LIKE '123'
+--Insert a new item by user_id 123
+INSERT INTO `items` SET `title`='do laundry', `details`='hand wash blouses', `timestamp`=now(), `user_id`='123'
+--Delete a item by user_id 123
+DELETE FROM `items` WHERE `title`='homework'
```
TimelikeClosure: Make sure you also add in the requirement that `` `user_id`=123``, otherwise you may accidentally delete an item by another user.

mysql_basics/mysql_queries.sql
```diff
@@ -1 +1,17 @@
-Place your queries below
+--Feature Set 1 - MySQL Basics
+SELECT * FROM `students` WHERE `username` LIKE 'yaerilim'
+UPDATE `students` SET `email`='helloyaeri@gmail.com' WHERE `username`='yaerilim'
+INSERT INTO `students` SET `name`='New Student', `email`='imnew@email.com', `age`='35', `SSN`='456789123', `username`='addedstudent', `password`=sha1('123newbee'), `joined`=now(), `status`= 'active'
+DELETE FROM `students` WHERE `username`='addedstudent'
+
+--Feature Set 2 - Make a Table
+--Select all todo_items made by user_id 123
+SELECT * FROM `items` WHERE `user_id` LIKE '123'
```
TimelikeClosure: - The table should be called `todo_items`, not `items`.
- While this works here, you'll want to get in the habit of leaving the quotes off values for numeric fields. For example, since the `` `user_id` `` field has a type  `int 11`, you'd just assign it a value of `123` instead of `'123'`.

mysql_basics/mysql_queries.sql
```diff
@@ -1 +1,17 @@
-Place your queries below
+--Feature Set 1 - MySQL Basics
+SELECT * FROM `students` WHERE `username` LIKE 'yaerilim'
```
TimelikeClosure: In MySQL syntax, the `LIKE` keyword is used for pattern matching (think RegEx), which is much slower than searching for an exact value. I'd suggest using an exact search unless you actually need to match a pattern.

PR #253, opened 2017-02-22
MySQL Basics - Andy Ong

Breathics:


mysql_basics/mysql_queries.sql
```diff
@@ -0,0 +1,23 @@
+Place your queries below
+
+-- Feature Set 1
+-- SELECT
+SELECT * FROM `students` WHERE `username`='andyong93'
```
TimelikeClosure: These should be referencing a table called `` `users` ``, not `` `students` ``.

